PR #5, opened 2015-02-11
Fix NullReferenceException when resubmitting a BrokeredMessage

xdansmith:
When you go to re-submit a message from in the DeadLetter queue, if the message contains a property that has a null value, a NullReferenceException would be thrown when the call is made to ToString()

Let me know if there are any questions,

Thanks!


paolosalvatori:
Thanks Dan!

PR #6, opened 2015-02-19
Handle null Message Property values when re-submitting deadletter msg

MarkHawcroft:
Handle null Message Property values when re-submitting deadletter messages


paolosalvatori:
Thanks Mark!

PR #7, opened 2015-02-25
fix for receiving messages error: Exception: Failed to lock one or more specified messages. The message does not exist.

greigs:


paolosalvatori:
Thanks Greig!

PR #13, opened 2015-07-31
Added option to create socket per task.

thebothead:
I have added an option to create a new socket per task.  Per the MSFT client-side scaling instructions creating a message factory per client is the best way to achieve this.  This is good for testing when you are saturating a single socket to the server and want to use more than one.


paolosalvatori:
Closing this pull request. Sorry, the product group made the same change, and it was already included in the source code. Thanks! :)

PR #17, opened 2015-09-07
Support byte array body (NServiceBus like messages)

SeanFeldman:
I'm trying to think of a way that would allow all NServiceBus users on ASB to use SB Explorer. 
NSB is using UTF8 encoded messages with BOM, stored as byte array.


SeanFeldman:
@paolosalvatori please review and let me know if this would suffice.
Also, any thoughts on making this part more pluggable?

SeanFeldman:
removed JSON formatting to ensure that XML is supported as well

PR #18, opened 2015-10-10
Read connection string from user environment variable AzureServiceBus…

SeanFeldman:
## Change

Allow reading of connection string from an environment variable scoped to the current user (`HKCU\Environment` `"AzureServiceBus.ConnectionString"`) to survive app.config resets

## Purpose of this change

Each time a new version is available, app.config is reset and changes are lost. It's easier to keep an environment variable to keep default connection out of app.config and safe from accidental commits.

### DIFF
--- a/Forms/MainForm.cs
+++ b/Forms/MainForm.cs
@@ -319,6 +319,7 @@ namespace Microsoft.WindowsAzure.CAT.ServiceBusExplorer
             eventClickFieldInfo = typeof(ToolStripItem).GetField(EventClick, BindingFlags.NonPublic | BindingFlags.Static);
             eventsPropertyInfo = typeof(Component).GetProperty(EventsProperty, BindingFlags.NonPublic | BindingFlags.Instance);
             GetServiceBusNamespacesFromConfiguration();
+            GetServiceBusNamespaceFromEnvironmentVariable();
             GetBrokeredMessageInspectorsFromConfiguration();
             GetEventDataInspectorsFromConfiguration();
             GetBrokeredMessageGeneratorsFromConfiguration();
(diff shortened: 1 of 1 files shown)


SeanFeldman:
@paolosalvatori please review. Thank you.

vonlar:
Is it also creating this environment variable?  Whenever I select "Enter connection string" it doesn't persist.

SeanFeldman:
> Allow reading of connection string

It was solely added for read-only purpose.

paolosalvatori:
You need to persist more than just a connection string. For example, subscriptionId and certificate thumbprint if you want to use SB metrics. Now, the maximum size for a user-defined environment variable is 32 KB. I suggest to serialize the entire config file into a single environment variable, so that you can read everything from it, not just a single connection string. Thoughts?

SeanFeldman:
Good suggestion. If serialization is done, it should support a way to modify that in a way users can do it themselves. In my case, I have a connection string that I rarely change. 

@paolosalvatori what if rather than loading from an environment variable a single namespace info, we'd specify a location for a (custom) configuration file that could contain multiple configurations? SBE would read the location from the env variable and process a file at the path specified as env variable value. That way we don't need to cramp it all into 32KB _and_ can have more than one namespace defined. Thoughts?

paolosalvatori:
Sounds good to me... the connection string could also start with http. In that case the configuration would be stored somewhere in the cloud (e.g. blob storage). It comes without saying, that the access to the file should be protected with a SAS token in this case.

SeanFeldman:
@vonlar would you like to raise a separate issue for that and possibly help out with implementation? 
Thanks.

vonlar:
@SeanFeldman Honestly I've been afraid to touch the connection string piece.  It might be easy to store, it might not be, but with work and a new baby on the way, I'm afraid I don't have the time right now to dive into it.  There's some info covered by the MVP NDA that needs to be brought to light.  Let me see if Paolo can share the information so I don't violate my NDA.

paolosalvatori:
@vonlar can you please send me an email and tell me more on "There's some info covered by the MVP NDA that needs to be brought to light"? Thanks :)

vonlar:
@paolosalvatori sent an email to your Hotmail account, but I need your Microsoft email address.  I was inaccurate, as it's covered by the Azure Advisor NDA, not the MVP NDA.

paolosalvatori:
My Microsoft account it's in the About form of the tool: paolos@microsoft.com :)

SeanFeldman:
I'm an advisor and MVP. Please CC on the thread or communicate separately
what the NDA item is. Thanks.

Issue #22, opened 2015-12-22
Queue EnableExpress is readonly when should be updateable


SeanFeldman:
`EnableExpress` is currently read-only, but should be possible to update.

Verification code:
```c#
var oldDesc = new QueueDescription("blah")
{
			EnableExpress = true,
};
		
await ns.CreateQueueAsync(oldDesc).ConfigureAwait(false);
	
var newDesc = new QueueDescription("blah") 
{
			EnableExpress = false,
};

await ns.UpdateQueueAsync(newDesc).ConfigureAwait(false);
```

paolosalvatori:
Hi Sean
could you make the change in my place? Otherwise I'll do after Xmas vacation, thanks!
Ciao
Paolo

SeanFeldman:
@paolosalvatori please review

PR #23, opened 2015-12-25
Allow EnableExpress to be updateable

SeanFeldman:


SeanFeldman:
Updated both [queue](https://github.com/paolosalvatori/ServiceBusExplorer/issues/22) & topic UI to handle change of EnableExpress setting.

@paolosalvatori please review

Issue #34, opened 2016-03-22
Re-submitting message with Guid property


jonsb:
When re-submitting a message that has a Guid property, the property is not included in the re-submitted message, apparently because the "value as string" cast below will return null (value is of type Guid, not string):

ConversionHelper.cs

public static object MapStringTypeToCLRType(string type, object value)
...
case "Guid":
                    return new Guid(value as string);
...

For me, changing "value as string" to "value.ToString()" seems to solve the problem, but I don't know the complete implications of this fix, and I don't know why "value as string" was written in the first place.

SeanFeldman:
A few thoughts:

1. Could have a unit test for this kind of functionality to seal what the expected behaviour should be (issue #36 is raised)
1. [`MapEDMTypeToCLRType`](https://github.com/paolosalvatori/ServiceBusExplorer/blob/7f87a15bbfcbb77715f2c9556b28cc0bb75f929b/Helpers/ConversionHelper.cs#L60) doesn't seem to be utilized at all

paolosalvatori:
Hi guys
any news on this issue?
Ciao
Paolo

SeanFeldman:
@jonsb would you like to submit a PR? If not, I can do it.

jonsb:
I'd be happy to submit a PR. But what about adding unit tests? I can add a unit test project and a couple of tests for my fix stating my assumptions. Ref issue #36, which is "done" - what does that mean?

SeanFeldman:
#36 is closed. My bad.

PR #37, opened 2016-04-07
Added support for "converting" a Guid object with MapStringTypeToCLRT…

jonsb:
…ype to Guid, and unit tests for this scenario. The reason for supporting the scenario is that other parts of the code calls this methods with a Guid object as parameter.


SeanFeldman:
Connects to #34

SeanFeldman:
:+1:

PR #54, opened 2016-10-09
Updated ServiceBus nuget package, targeted AnyCPU, targeted .Net 4.6.1

vonlar:
Changes made:

- I've updated the ServiceBus nuget package.  
- Created a new Configuration for AnyCPU
- Changed the output to AnyCPU 
- Removed the preference for 32 bit
- Updated the target .Net framework to 4.6.1

Changes not made:
- Increment version number
- Have the MSI install to 'Program Files' instead of 'Program Files (x86)'


SeanFeldman:
@paolosalvatori you should have increased the version as it's not the same 3.4 as it was before 😉

paolosalvatori:
I realized that and increased the version right yesterday to 3.4.0 :)

vonlar:
Any way to get the installer to put it in "Program Files" instead of "Program Files (x86)"?

paolosalvatori:
I don't know... I didn't make the installer :-)

PR #55, opened 2016-10-18
Keep peeking queue and subscription deadletter messages until the requested count has been peeked.

joellyons:
A maximum of 250 messages were being peeked when viewing the deadletter messages from a queue or subscription. This change uses the looping approach used elsewhere in SBE to continue until the requested count has been peeked.

### DIFF
--- a/Controls/HandleQueueControl.cs
+++ b/Controls/HandleQueueControl.cs
@@ -1475,15 +1475,22 @@ namespace Microsoft.WindowsAzure.CAT.ServiceBusExplorer
                     var queueClient =
                         serviceBusHelper.MessagingFactory.CreateQueueClient(
                             QueueClient.FormatDeadLetterPath(queueDescription.Path), ReceiveMode.PeekLock);
-                    var messageEnumerable = queueClient.PeekBatch(count);
-                    if (messageEnumerable == null)
+                    var totalRetrieved = 0;
+                    int retrieved;
+                    do
                     {
-                        return;
-                    }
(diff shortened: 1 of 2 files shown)


SeanFeldman:
Connects to #56

Controls/HandleQueueControl.cs
```diff
@@ -1475,15 +1475,22 @@ private void GetDeadletterMessages(bool peek, bool all, int count, IBrokeredMess
...
+                            : count - totalRetrieved);
+                        var enumerable = messages as BrokeredMessage[] ?? messages.ToArray();
+                        retrieved = enumerable.Count();
+                        if (retrieved == 0)
+                        {
+                            continue;
+                        }
+                        totalRetrieved += retrieved;
+                        brokeredMessages.AddRange(messageInspector != null ? enumerable.Select(b => messageInspector.AfterReceiveMessage(b, writeToLog)) : enumerable);
+                    } while (retrieved > 0 && (all || count > totalRetrieved));
```
SeanFeldman: Should 

> `while (retrieved > 0 && (all || count > totalRetrieved));` ?

be

> while (retrieved > 0 && (all || count >= totalRetrieved));

If someone requested 20 messages, and 20 where received, then stop?

joellyons: I think if we use >= then it will loop again and get more results, which we don't want because if count == totalRetrieved then we've retrieved all that was requested and we need to stop looping.  So, we only want to keep looping if the totalRetrieved is less than the count requested.  This is the pattern used in many places in this project.

SeanFeldman: fair point

PR #149, opened 2017-09-26
Async delete

danielmarbach:
## Before

https://youtu.be/Nl82ePL8nLE

## After

https://youtu.be/qSw5AHzIUyw

The video is still processing. But the TLDR version is that deleting is with my example here is almost one minute faster.

### DIFF
--- a/src/ServiceBusExplorer/Helpers/ServiceBusHelper.cs
+++ b/src/ServiceBusExplorer/Helpers/ServiceBusHelper.cs
@@ -2397,7 +2397,7 @@ namespace Microsoft.Azure.ServiceBusExplorer
         /// Deletes the queue passed as a argument.
         /// </summary>
         /// <param name="queueDescription">The queue to delete.</param>
-        public void DeleteQueue(QueueDescription queueDescription)
+        public async Task DeleteQueue(QueueDescription queueDescription)
         {
             if (queueDescription == null)
             {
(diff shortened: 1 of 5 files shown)


SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.0.104-PullRequest0149.9.build.297 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/build/4.0.104-PullRequest0149.9.build.297) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/dc33ef24a1 by @danielmarbach)

danielmarbach:
I work with sean, if he would he would be in serious trouble ;)

danielmarbach:
I do think there is probably more cleanup and adjustment necessary but this is a first step in the right direction and if I find some spare time I might be able to send in a few more PRs (nothing promised though)

src/ServiceBusExplorer/Helpers/ServiceBusHelper.cs
```diff
@@ -1162,16 +1162,13 @@ public void DeleteRelay(string path)
...
         {
             if (relayServices == null)
             {
                 return;
             }
-            foreach (var relayService in relayServices)
-            {
-                DeleteRelay(relayService);
-            }
+            await Task.WhenAll(relayServices.Select(DeleteRelay).ToArray());
```
SeanFeldman: nit: `.ToArray()` is not needed.

SeanFeldman:
Removed .ToArray() [here](https://github.com/paolosalvatori/ServiceBusExplorer/pull/150/files#diff-f8fb4ec2c6ff293b7f32aba3d557984e).

SeanFeldman:
> I work with sean, if he would he would be in serious trouble ;)

I've contemplated to piss off @danielmarbach and ask to suffix methods in a code base that is not entirely async... But then he'd not leave me alone 😛 

To be fair, the whole async suffix discussion is a mood point. At some point Clemens joked about suffixing the non-async code with a snail emoji (can't find the tweet). I wouldn't mind that 😄

paolosalvatori:
Thanks @danielmarbach for the improvement :) Keep sending :) Indeed the tool still contains non-async method calls, as it was written before the await/async mechanism was added. Some parts were converted to await/async, but not all of them.

PR #180, opened 2018-01-28
More reliable and faster purging of messaging entities, fix for #60

### LINKED ISSUE #60: Recieve and Delete large amount of messages does not work
If you choose to Receive and Delete 50,000 messages from a queue/subscriptions/deadletter it will typically only 2000 - 4000 messages, instead of the specified amount. The amount of messages that gets removed is different each time; sometimes it's 2900 messages, the next time it might be 3600. Doing Receive and Delete All has the same problem as well. I'm not sure if the tool is timing out or how it determines when to stop receiving messages, but it seems it does hit some limit. This makes it tough to clear out our deadletter queue when it has a 200,000 messages in it. We've written a console app to do this for us, but it would be great if it just worked inside of Service Bus Explorer.

ErikMogensen:
Attempts to fix #60. There are many variations with partioned entities, session entities, message size, network latency and other factors so the code will probably have to be tuned.

I used VS automatic formatting and it did a lot of formatting on other places too. I thought I'd leave those changes since I think it looks better and will take a while to undo. 

I had some issues with entities with sessions so only one receiving task is used for them. I upgraded to the latest Service Bus client, but it didn't help. But that is how I have tested this so the code upgrades the Service Bus client to the latest version.

I moved code from the HandleQueueControl and the HandleSubscriptionControl classes to a new class called MessagingPurger where subscriptions and queues are handled in basically the same way.

### DIFF
--- a/src/ServiceBusExplorer/Controls/HandleQueueControl.cs
+++ b/src/ServiceBusExplorer/Controls/HandleQueueControl.cs
@@ -345,8 +345,8 @@ namespace Microsoft.Azure.ServiceBusExplorer.Controls
             "SharedAccessKey"
         };
 
-        private static readonly List<string> Operators = new List<string> {"ge", "gt", "le", "lt", "eq", "ne"};
-        private static readonly List<string> TimeGranularityList = new List<string> {"PT5M", "PT1H", "P1D", "P7D"};
+        private static readonly List<string> Operators = new List<string> { "ge", "gt", "le", "lt", "eq", "ne" };
+        private static readonly List<string> TimeGranularityList = new List<string> { "PT5M", "PT1H", "P1D", "P7D" };
 
         #endregion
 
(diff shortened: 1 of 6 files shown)


SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.1.0-PullRequest0180.162.build.344 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/build/4.1.0-PullRequest0180.162.build.344) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/9995d0dfb6 by @ErikMogensen)

SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.1.0-PullRequest0180.163.build.345 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/build/4.1.0-PullRequest0180.163.build.345) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/9ba5f5119b by @ErikMogensen)

paolosalvatori:
Thanks a lot @ErikMogensen!!!

ErikMogensen:
Thanks for providing this great tool, Paolo!

PR #184, opened 2018-02-03
Purging DLQs with a single message

### LINKED ISSUE #183: Purge DLQ is broken
SBE version: 4.0.106
Queue: not partitioned, standard tier, no sessions, 1GB size

Looks like there was a regression in #180.
As of now, Purge DLQ doesn't do anything.

Validation: have a single dlq-ed message, push Purge DLQ. Status says:
><22:28:57> [0] messages have been purged from the deadletter queue of the [test] queue in [896] milliseconds.

and dead-lettered message stays as is.
image

SeanFeldman:
Fixes #183 

I took the liberty to clean up the class a little bit and remove some redandencies.
Couldn't figure out what to do about https://github.com/paolosalvatori/ServiceBusExplorer/commit/63e6c9d67f14e31b54d0e7ee12fbff4447a07d4e. @paolosalvatori should that method be entirely removed as it was (and is not) ever called?

### DIFF
--- a/src/ServiceBusExplorer/Controls/HandleQueueControl.cs
+++ b/src/ServiceBusExplorer/Controls/HandleQueueControl.cs
@@ -453,7 +453,6 @@ namespace Microsoft.Azure.ServiceBusExplorer.Controls
 
         public async Task<long> PurgeDeadletterQueueMessagesAsync()
         {
-            var dlqPath = QueueClient.FormatDeadLetterPath(queueDescription.Path);
             using (var deleteForm = new DeleteForm($"Would you like to purge the deadletter queue of the {queueDescription.Path} queue?"))
             {
                 if (deleteForm.ShowDialog() != DialogResult.OK)
--- a/src/ServiceBusExplorer/Helpers/MessagingPurger.cs
+++ b/src/ServiceBusExplorer/Helpers/MessagingPurger.cs
@@ -325,16 +325,20 @@ namespace Microsoft.Azure.ServiceBusExplorer.Helpers
         private bool EntityIsPartioned()
         {
             if (queueDescription != null)
+            {
                 return queueDescription.EnablePartitioning;
-            else
-                return subscriptionWrapper.TopicDescription.EnablePartitioning;
+            }
+
+            return subscriptionWrapper.TopicDescription.EnablePartitioning;
         }
         private bool EntityRequiresSession()
         {
             if (queueDescription != null)
+            {
                 return queueDescription.RequiresSession;
-            else
-                return subscriptionWrapper.SubscriptionDescription.RequiresSession;
+            }
+
+            return subscriptionWrapper.SubscriptionDescription.RequiresSession;
         }
         #endregion
     }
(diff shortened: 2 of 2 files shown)


SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.1.0-PullRequest0184.172.build.352 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/build/4.1.0-PullRequest0184.172.build.352) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/7c6b3c1e40 by @SeanFeldman)

ErikMogensen:
Thanks for fixing this, @SeanFeldman . I wrote the MessagePurger class and a lot of it is based on code moved from other classes in in SBE. I see that the method EntityIsPartioned is used by a variable, but the variable is not used so go ahead and delete the method and the variable called partioned. BTW, how did you find that the variable wasn't used?

I agree with most of your changes. However, I do not agree about the removal of the #regions since it looks like most classes in SBE are using regions in that way. Thus it would make MessagePurger.cs inconsistent with the other files in SBE.

SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.1.0-PullRequest0184.173.build.356 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/build/4.1.0-PullRequest0184.173.build.356) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/596ea464fc by @SeanFeldman)

SeanFeldman:
@ErikMogensen while refactoring the code the variable was marked as unused. Once I've removed it, the method was orphaned. And it's a private method, meaning no external calls. I've compared it to the original code and there was nothing else calling it.

As for the #regions - w/o being disrespectful to anyone, it feel a bit as a cargo cult. `MessagePurger` is a self contained class that doesn't really require to be divided in regions. As in regards to consistency with other files of SBE - I always ask the following question: If everyone jumps off a roof, would you jump as well? Similar story, rather than making new code follow the guidelines of something that is not necessarily either up-to-date or appropreate, shouldn't we strive to modernize things? Using `var` rather than explicit types where type is either infered from the variable name or not important. Using `out var _` rather than explicitely defining a temporary variable. Using lambdas rather than defining function callbacks. All these are the improvements we make over the time and not always on the entire code base.

And if that was not convincing enough, here's an example, [private member fields](https://github.com/paolosalvatori/ServiceBusExplorer/pull/184/files#diff-96ddc364bbbf9669c3d6aa5bb1784352L38). Why would you need a region around three private member variable to say those are private fields? It just adds noise. Imagine I add a constant there accidentally, now it forces me to move it out and do more gold plating work that I should.  Or [public constructions region](https://github.com/paolosalvatori/ServiceBusExplorer/pull/184/files#diff-96ddc364bbbf9669c3d6aa5bb1784352L45). Why would I need a region around those if they are obviously constructors? 🙂 You get the idea.

A class (ideally) should be small and contain a scoped / single responsibility to make its purpose clear. When it's unclear what a class is doing, regions kick in to compensate, but not to fix the problem. SBE has gone a loooong way, and it's not a simple code base. Though I can bet if it was repaved, the need for regions at a lot of places would be gone. Hope that answers the rational behind my decision to remove it.

ErikMogensen:
@SeanFeldman , thank you for your thorough answer. I agree that sometimes it is better to do changes gradually even if it makes the code base inconsistent. 

Using #regions cause extra lines of code, but on the other hand it enforces a certain structure for the source files which is good in an open source project. Also, some people may think it is easier to navigate in a class if you can collapse sections of it. To me it seems to be a question of personal taste and we should have @paolosalvatori decide whether he wants #regions in all classes or just the big ones.

paolosalvatori:
I love #regions, they allow me to organize code in subsections. I'd prefer to keep them.

Issue #173, opened 2017-12-20
Repair and resubmit all messages from DLQ


antongeorgiev:
It is not possible to repair and resubmit large amount of messages from the dead-letter queue because currently messages can only be resubmitted in a batch mode and a batch size is limited to 256 KB.

A user should be able to either select the queue or manually select multiple or all messages in a queue and choosing to "Repair and resubmit in batch mode" should have all selected messages processed.

Possible solution might be to leverage the functionality in the "Send messages" dialog by using the DLQ as an input source and providing an option to select the destination queue but defaulting to the entity the DLQ is part of. 

The following posts might be helpful:

- https://weblogs.asp.net/sfeldman/asb-batching-brokered-messages
- https://mikhail.io/2017/07/sending-large-batches-to-azure-service-bus/

antongeorgiev:
The following posts might be helpful:

- https://weblogs.asp.net/sfeldman/asb-batching-brokered-messages
- https://mikhail.io/2017/07/sending-large-batches-to-azure-service-bus/

SeanFeldman:
@antongeorgiev it's not possible to repair and resubmit more than a single message. SBE was designed to handle only one message at a time and it is not related to a batch size. The current UI would need to be modified and better not reuse-abuse existing functionality as it becomes more and more difficult to maintain this code based.

I don agree that having an ability to fix multiple messages from DLQ and resubmit would be nice to have. Would you like help and send a PR?

paolosalvatori:
Hi @antongeorgiev one of the articles you sent me refer my Service Bus Extensions library that you find on this GitHub account ;) The reason why the tool is designed this way it's because:
 - if you want to resubmit multiple messages, you can select them and send them as is, with no changes
 - if you want to repair and resubmit messages, there's no wayout, you have to do this operation one message at the time.
Hope this helps. You have another idea in mind, please let us know.

ErikMogensen:
@antongeorgiev, in the latest version of the Service Bus Explorer (SBE) in the Develop branch, there is new functionality for submitting multiple messages from the DLQ. There is a new checkbox with the text "Remove messages from the DLQ". If you check it, SBE will send the messages one at a time to avoid duplicates.

Is that what you are looking for? 

Anyway, I think that the "Resubmit selected messages in batch mode" should be changed to just send as many messages that fits in a batch and then create more batch messages to be able to send all of the selected messages. Paolos Service Bus Extensions Library seems great for this task.

[...]

Issue #135, opened 2017-07-04
Need to complete the message repair and re-transmit pattern


paolosalvatori:
Need to complete the message repair and re-transmit pattern:
 - Defer the original message
 - Receive the deferred message by sequence number to consume it

Repair and resubmit is already in place. What's missing is the safe original message removal leveraging steps above.

SeanFeldman:
Similar concept implemented by SB360 with ["Deferred Dead Letter" feature](https://www.servicebus360.com/blogs/resubmit-delete-messages-azure-service-bus-messaging/).

PR #192, opened 2018-03-04
Enable the message retransmit pattern for messages in DLQ

ErikMogensen:
This change partially solves #135. It only handles messages in deadletter subqueues but not in normal queues and subscriptions.

The reason for this is that it is not possible to call Defer() on a message unless it has been acquired using one of the Receive* methods, so it won’t work if Peek* has used to get the message. MikeWo explains the reason for this in his answer on [StackOverflow](https://stackoverflow.com/questions/33620999/complete-a-message-in-a-dead-letter-queue-on-azure-service-bus).

One way to solve this is by iterating through the messages in the queue using a Receive* method in PeekLock mode and when the message to be moved is found, it is Cloned, Sent and Completed. When this is done all the received messages are Abandoned and Disposed. Abandoning a message on the main queue causes its Delivery count to increment, which in turn might cause messages to be moved to the DLQ. On the other hand, abandoning messages on DLQ should not have any bad effect.

I therefore implemented the above for only DLQ. The DLQ is where I see the big need for this feature.
The timespan for doing the receiving and the send must be less than the Lock duration for the queue which may be a problem in cases with many or large messages.
 
If this works fine it could be implemented for the main queue and for the Transfer Deadletter Queue, with some kind of warning that the Delivery Count will be increased.

For the Resubmit form I added a check box to remove the message from the DLQ. When it is checked it handles one message at a time to make sure that no messages are lost, unless they have the same MessageID and the receiving queue/topic has Message Duplication check turned on. This is different behavior from when the check box is not checked, because then it tries to put all the messages in the same batch. I believe that is the root cause for issue described in #173 but I did not fix it for the code path when it is not removing messages.

I changed the form for selecting target queue/topic to default to select the main queue/topic for the DLQ since I know that behavior have been requested and I got tired of selecting the queue/topic every time I tested submitting messages.

I also added a menu item for deleting selected messages from the DLQ when looking at them in the grid.

This PR is a big modification and there are many different combinations of message size, network latency, partitioning, queue/topic/subscription settings, Service Bus health, Service Bus SKU and I haven’t tested them all so there is a risk that this change might cause bugs.


SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.0.108-PullRequest0192.30.build.368 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/build/4.0.108-PullRequest0192.30.build.368) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/6122922a2e by @ErikMogensen)

paolosalvatori:
Thanks @ErikMogensen, great work. I did the same analysis a few weeks ago and I came to the same conclusion regarding messages in normal queues and topics. Thanks for the great addition, you are definitely one of the most valuable contributors to this tool :)

PR #206, opened 2018-04-23
Send messages one by one when using Repair and Submit in Batch Mode

### LINKED ISSUE #173: Repair and resubmit all messages from DLQ

ErikMogensen:
I believe this fixes #173.

This change makes SBE send messages one by one instead of trying to send all the selected messages in a batched message. The reason I made this change is that it will now allow a lot of messages to be submitted, instead of being limited to the maximum message size which currently is 256 KB or 1 MB depending upon SB SKU. The only disadvantage I see is that the risk of not sending all messages increases. Therefore SBE will now display a MessageBox if not all messages were sent.

### DIFF
--- a/src/ServiceBusExplorer/Forms/MessageForm.cs
+++ b/src/ServiceBusExplorer/Forms/MessageForm.cs
@@ -489,14 +489,13 @@ namespace Microsoft.Azure.ServiceBusExplorer.Forms
                                     await messageSender.SendAsync(outboundMessages[messageIndex++]);
                                 }
                             }
-                            catch (Exception)
+                            catch (Exception exception)
                             {
                                 Application.UseWaitCursor = false;
-                                string messageText = $"Only sent {messageIndex} messages when {outboundMessages.Count}" +
-                                    " messages were selected.";
+                                string messageText = $"{outboundMessages.Count} were selected but only" +
+                                    $" {messageIndex} messages were sent. The error message is: {exception.Message}";
                                 MessageBox.Show(messageText, "Not all selected messages were sent",
                                     MessageBoxButtons.OK, MessageBoxIcon.Exclamation);
-
                             }
 
                             stopwatch.Stop();
--- a/src/ServiceBusExplorer/Forms/MessageForm.cs
+++ b/src/ServiceBusExplorer/Forms/MessageForm.cs
@@ -481,7 +481,24 @@ namespace Microsoft.Azure.ServiceBusExplorer.Forms
                         }
                         else
                         {
-                            await messageSender.SendBatchAsync(outboundMessages);
+                            var messageIndex = 0;
+                            try
+                            {
+                                while(messageIndex < outboundMessages.Count)
+                                {
+                                    await messageSender.SendAsync(outboundMessages[messageIndex++]);
+                                }
+                            }
+                            catch (Exception)
+                            {
+                                Application.UseWaitCursor = false;
+                                string messageText = $"Only sent {messageIndex} messages when {outboundMessages.Count}" +
+                                    " messages were selected.";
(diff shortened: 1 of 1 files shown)


SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.1.0-PullRequest0206.207.build.387 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/build/4.1.0-PullRequest0206.207.build.387) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/7dea00d74a by @ErikMogensen)

paolosalvatori:
We may also send messages in batches using this extension https://github.com/paolosalvatori/ServiceBusExtensions or using a similar plug-in made by @SeanFeldman

ErikMogensen:
@SeanFeldman, I choose to not implement batching since it was complicated and I considered it would solve the issue without that. I do think adding batching is probably a good idea, but it requires a lot of more effort to implement and test. 

I don't think batching messages using SendBatch/SendBatchAsync saves any money. I could not find any documentation regarding that though. It is mentioned in https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-performance-improvements that two other types batching do not affect cost.

SeanFeldman:
> I don't think batching messages using SendBatch/SendBatchAsync saves any money.

I'll stand corrected, it is not. So I'll leave that point until the answer is available. The other concerns that I had/have were performance and UX related. Sending 100 individual messages vs sending a single batch has a difference. [Client-Side batching](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-performance-improvements#client-side-batching) is a mechanism to somewhat helps with the performance associated with individual sends, but it can go so much. 

The minimum we could do on this specific PR is to collect the tasks and execute them all rather than await one-by-one. That alone would be more performant compared to single awaited sends.

To address any potential work - if we're in agreement that this is not fully fleshed out, we should have created a follow up/todo issue? We've knowingly degraded the performance and experience w/o leaving any breadcrumbs to action later. That feels off to me.

ErikMogensen:
I agree that it should be a todo issue. If the sends were executed in parallell, message ordering would be lost, which is also a change in behavior. I therefore think the best thing would be to just add message batching.

SeanFeldman:
The issue reminds me of a salesman shipwrecked on a tropical island joke... 🙂

ErikMogensen:
@SeanFeldman , I curious about that joke, I don't recognize it...

SeanFeldman:
A salesman shipwrecked on a tropical island of an unbelievable beauty untoched by the modern society. To his surprise, he sees a local laying on the beach, clapping his hands and a little monkey jumps of a parltree with a coconut. Cracks it open, inserts a straw and serves it.

The salesman guy goes bullistic, runs to the local, shakes him up and yells. "Dude, dude! Are you crazy?!". The local takes off shades and wondering "what?!". The sales guy goes "we could get investments, start a resort constructions, bring tourists, make a ton of cash, retrire early, and sit on the beach doing nothing". The local looks at him and goes "and what am I doing now?..."

Back to the batching.

Issue #186, opened 2018-02-11
Change C# language version


ErikMogensen:
Currently the Language version is set to "C# latest major version (default)". I suggest this is changed to "C# latest minor version (latest)". That way we could take advantage of the new features in C# 7.1, 7.2 and later versions.

Do you agree?

SeanFeldman:
Sounds good. Any specific features you're thinking of?

ErikMogensen:
I was thinking of [Reference semantics with value types](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-7-2#reference-semantics-with-value-types) and [Non-trailing named arguments](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-7-2#non-trailing-named-arguments). 

It would also be nice with the C# 7.0 support for tuples but as I understand it that requires an upgrade to .NET 4.7 or having to include one more dll when distributing SBE which may be not worth it.

SeanFeldman:
Tuples require an external [package](https://www.nuget.org/packages/System.ValueTuple/)

ErikMogensen:
It was only changed for Debug builds, not for Release builds.

paolosalvatori:
Let me check to understand the implications in terms of version of the .NET framework and Visual Studio.

ErikMogensen:
@paolosalvatori, I tried different combinations of the C# version setting and the Target framework setting and some combinations do not compile. It looks like the target version of the .NET Framework is solely determined by the Target framework setting (on the Application tab in project settings).

paolosalvatori:
@ErikMogensen which version should I set? :)

paolosalvatori:
Ok I approved the PR, I need @SeanFeldman to approve it as well :)

SeanFeldman:
`LangVersion` version has no impact on the compiled assembly. It's a syntactic sugar.

> Regardless of which -langversion setting you use, you will use the current version of the common language runtime to create your .exe or .dll. 

For that reason I'm 👎 to merge that change.

ErikMogensen:
I'll try to convince you.

By setting the C# language to a fixed version you know which features you may use and not use when developing. 

[...]

PR #258, opened 2018-09-23
Set C# version to 7.3

ErikMogensen:
Fixes the release builds part of #186.


SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.1.0-PullRequest0258.273.build.473 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/build/4.1.0-PullRequest0258.273.build.473) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/424c5bf0af by @ErikMogensen)

SeanFeldman:
Yes, it should be always set.
Is it still not possible to use the new csproj with winforms? Would make csproj files so much cleaner.

SeanFeldman:
@paolosalvatori PR pending your review

ErikMogensen:
I built the version of SBE that was merged today with PR #261, which is built with C# 7.3, and tested it on a Windows machine running .NET 4.6.2 and there were no issues. 

I used this PowerShell expression to get .NET version:
Get-ChildItem "HKLM:SOFTWARE\Microsoft\NET Framework Setup\NDP\v4\Full\" | Get-ItemPropertyValue -Name Release 

and used the table at https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/how-to-determine-which-versions-are-installed#to-find-net-framework-versions-by-querying-the-registry-in-code-net-framework-45-and-later to interpret the output.

But this PR is obsolete now since #263 already incorporates this :)

ErikMogensen:
Bringing up this one again since I was contacted by a friend who was unable to build SBE. It turned out to be that he needed to update his Visual Studio. Even though the readme says that you need a minimum version of VS 2017, most people don't read it. 

I believe that there are many people who try to build SBE, but give up when they get strange build errors.

SeanFeldman:
What do you suggest?

ErikMogensen:
I suggest we set C# version to 7.3.

SeanFeldman:
Isn't `latest` is 7.3?

>  Projects targeting any released version will use C# 7.3 (the latest released version). This behavior means that any project targeting .NET Framework will use latest (C# 7.3).

_[source](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version)_

ErikMogensen:
Yes, 7.3 is the latest supported version with 8 in preview. By explicitly setting C# to 7.3 then you get an error telling you that your VS to old if you try to build SBE with a an old version of VS.

SeanFeldman:
I'm not a big fan of constraining the latest version just to generate an
eerie message for those that use an old version of VS.

Ignorance of law excuses no one.

ErikMogensen:
It is not about constraining. It is about being clear what the constraints/law is. Otherwise how shall people know what is the expected version of C# to use? The latest may not always be the right answer.

SeanFeldman:
I question if having 7.3 version hard-coded in the csproj would make it more clear. Regardless of what I think, if decision will be to change it and update each time latest is a different version, by all means 🙂

ErikMogensen:
I don't mind doing creating PRs for that. I think it will be once or twice per year.

ErikMogensen:
The hard work is getting the PRs approved ;)

SeanFeldman:
The hardship is in the nature of async discussions.
I tend not to approve w/o having to agree on what's changed or having expectations sorted out.
Yes, I'm old 🙂

ErikMogensen:
:)

Issue #235, opened 2018-07-11
Put all the all the appsettings from the config file in a personal user file


ErikMogensen:
Currently all appSettings are overwritten if you download a new version of ServiceBus Explorer. Also these settings are shared between all users running ServiceBus Explorer from the same directory, which may be many if it is located on a shared location.

To fix this these settings should be stored in a file in the users personal folder under the roaming profile. To keep backwards compability these settings should also be read from the application config file, but the settings in the user file will override those. 

This may create confusion for someone changing the application config file manually and expecting the changes to take effect. To mitigate that the complete path to the user file should be visible in the Options dialog window with a label "Settings file (overrides application config file settings):" .

This is a requirement to make #232 work nicely since there will be a new setting for the location of connection strings.

ErikMogensen:
Fixed in #261

Issue #217, opened 2018-05-17
Enable WIP plugin


SeanFeldman:
I have no permission to access settings, so @paolosalvatori ping #2.

https://github.com/paolosalvatori/ServiceBusExplorer/issues/212#issuecomment-389561214

paolosalvatori:
@SeanFeldman @tomkerkhove  Should be ok now. I installed the plugin in all my repos. Can you please check?

PR #261, opened 2018-09-30
Enable having connection strings and settings in a user config file

### LINKED ISSUE #232: Put the connection strings in a file in the users personal folder
That way the strings would be a little bit more protected and when downloading a new version of SBE you’d keep your connection strings. 

Add an option to use the connection strings in the application config file for backwards compatibility, even though it is not recommended. 

In the future encryption of the strings may be added using DPAPI and that may not work when using the Roaming profile. Therefore use the local non-roaming profile for this.

ErikMogensen:
Fixes #232 and #235. 

If the user configuration file does not exist it will automatically be created when a connection is added or a setting is changed. The user configuration file will be put in the Roaming folder. By default it is called  C:\Users\<username>\AppData\Roaming\Service Bus Explorer\UserSettings.config. I do not see a security issue having the connection strings stored as part of the roaming profile instead of in the directory where SBE is running from. But if anyone see a problem with this, please let me know.

All changes to the settings are done to the user configuration file. There are still some actions that would need SBE to run in elevated mode if it is installed under the Program Files (x86) folder since Windows by default protects all writing to files under this folder. The actions that I am aware of that still need administrative privileges are changes to connection strings stored in the application config file and changes to the message file.

When a setting exist in both the user configuration file and the application configuration file the setting from the user configuration file will be used.

This PR removes all writing to the app.config file, which should lessen the risk of committing unwanted changes to that file.

It fixes some minor bugs that needed to be fixed to be able to test some parts of the PR. Some parts of the MainForm class were moved to other classes located in the Helpers folder to be able to call them from other parts of the application. This has the added benefit of decreasing the size and complexity of the still huge MainForm class.

It requires C# 7.1 support so there is a dependency on #186. To be able to build the projects the change in C# language is part of this PR.


SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.1.0-PullRequest0261.293.build.476 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/build/4.1.0-PullRequest0261.293.build.476) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/e061028991 by @ErikMogensen)

SeanFeldman:
> having the connection strings stored as part of the roaming profile

👍 If this will be an issue, we'll hear from users.

src/ServiceBusExplorer/Helpers/Constants.cs
```diff
@@ -31,6 +31,13 @@ namespace Microsoft.Azure.ServiceBusExplorer.Helpers
...
+        public const string QueueEntities = "Queues";
+        public const string TopicEntities = "Topics";
+        public const string RelayEntities = "Relays";
+        public const string EventHubEntities = "Event Hubs";
+        public const string NotificationHubEntities = "Notification Hubs";
+
+
         public static readonly TimeSpan DefaultOperationTimeout = TimeSpan.FromMinutes(1.0);
         public static readonly TimeSpan TokenRequestOperationTimeout = TimeSpan.FromMinutes(3.0);
         public static readonly long ServicePointMaxIdleTimeMilliSeconds = 0xc350;
```
SeanFeldman: Unrelated, but funny... `0xc350` is better than `50000` 😂

src/ServiceBusExplorer/Helpers/DoubleHelper.cs
```diff
@@ -0,0 +1,52 @@
...
+// KIND, EITHER EXPRESS OR IMPLIED. SEE THE LICENSE FOR THE SPECIFIC LANGUAGE GOVERNING 
+// PERMISSIONS AND LIMITATIONS UNDER THE LICENSE.
+//=======================================================================================
+#endregion
+
+using System;
+
+namespace Microsoft.Azure.ServiceBusExplorer.Helpers
+{
+    public class DoubleHelper
```
SeanFeldman: nit: I tend to call extension methods container class `{extended-subject}Extensions`. So something like `DoubleExtensions`.

src/ServiceBusExplorer/Helpers/EventProcessorCheckpointHelper.cs
```diff
@@ -82,7 +82,7 @@ public static void ReadCheckpoints()
         /// </summary>
         public static void WriteCheckpoints()
         {
-            if (itemList.Count == 0)
+            if (itemList == null || itemList.Count == 0)
```
SeanFeldman: Since we're on the latest version of C#, can do

> (items?.Count == 0)

src/ServiceBusExplorer/Helpers/MessageAndPropertiesHelper.cs
```diff
@@ -51,6 +51,7 @@ public static class MessageAndPropertiesHelper
         private const string Key = "Key";
         private const string Type = "Type";
         private const string Value = "Value";
+        private const string DefaultMessageText = "Hi mate, how are you?";
```
SeanFeldman: 😄

src/ServiceBusExplorer/Helpers/MessageAndPropertiesHelper.cs
```diff
@@ -287,6 +282,39 @@ public static List<MessagePropertyInfo> ReadProperties()
...
+                        var root = XElement.Load(xmlReader);
+                        var cdata = root.DescendantNodes().OfType<XCData>().FirstOrDefault();
+                        if (cdata != null)
+                        {
+                            return cdata.Value;
+                        }
+                    }
+                }
+            }
+            // ReSharper disable once EmptyGeneralCatchClause
```
SeanFeldman: I'm not 100% sure, but think if you comment inside `catch` block, R# will not complaint.

src/ServiceBusExplorer/Helpers/MessageAndPropertiesHelper.cs
```diff
@@ -287,6 +282,39 @@ public static List<MessagePropertyInfo> ReadProperties()
...
+                    return null;
+                }
+
+                using (var reader = new StreamReader(messageFilePath))
+                {
+                    using (var xmlReader = XmlReader.Create(reader))
+                    {
+                        var root = XElement.Load(xmlReader);
+                        var cdata = root.DescendantNodes().OfType<XCData>().FirstOrDefault();
+                        if (cdata != null)
```
SeanFeldman: If  `cdata` is null, we return a null. Therefore could do straight

`return cdata?.Value`

src/ServiceBusExplorer/Helpers/XmlLinqExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+
+#region Using Directives
+
+using System.Xml.Linq;
+
+#endregion
+
+namespace Microsoft.Azure.ServiceBusExplorer.Helpers
+{
+    public static class XmlLinqExtensions
```
SeanFeldman: 👍 for name

src/ServiceBusExplorer/Helpers/XmlLinqExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+    {
+        public static XElement AquireElement(this XContainer container, 
+            string name, bool addFirst = false)
+        {
+            var element = container.Element(name);
+
+            if (null == element)
```
SeanFeldman: I'd invert. Firs, check if `element` is _not_ null and return immediately. Everything else doesn't have to be indented and in a block then.

src/ServiceBusExplorer.Tests/Helpers/TwoFilesConfigurationTests.cs
```diff
@@ -0,0 +1,906 @@
...
+using System.IO;
+using System.Xml;
+using System.Xml.Linq;
+
+using Microsoft.Azure.ServiceBusExplorer.Helpers;
+using Microsoft.ServiceBus;
+
+using NUnit.Framework;
+
+#endregion
```
SeanFeldman: No need in regions for tests.

src/ServiceBusExplorer.Tests/app.config
```diff
@@ -18,6 +63,39 @@
...
+        <assemblyIdentity name="Microsoft.Azure.Services.AppAuthentication" publicKeyToken="31bf3856ad364e35" culture="neutral" />
+        <bindingRedirect oldVersion="0.0.0.0-1.0.3.0" newVersion="1.0.3.0" />
+      </dependentAssembly>
     </assemblyBinding>
   </runtime>
-<startup><supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.6.1" /></startup></configuration>
+  <startup>
+    <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.6.1" />
+  </startup>
+  <system.serviceModel>
```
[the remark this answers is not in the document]
ErikMogensen: Looking at https://stackoverflow.com/questions/21566528/what-happens-if-i-remove-the-auto-added-supportedruntime-element I think it is better to keep it. Better to get an error immediately than a runtime exception when executing it. It is vnlikely that this will ever be an issue in a dev or test environment but having spent some time with a failing build in AppVeyor I appreciate all checks of making sure prerequisites are installed.

src/ServiceBusExplorer/Helpers/DoubleHelper.cs
ErikMogensen: I totally agree. It didn't start out as an extension method.

src/ServiceBusExplorer/Helpers/EventProcessorCheckpointHelper.cs
ErikMogensen: Sure.

src/ServiceBusExplorer/Helpers/XmlLinqExtensions.cs
ErikMogensen: Ok. I changed it, but I preferred the original way with only one exit from the method. But no big deal for me :)

src/ServiceBusExplorer.Tests/Helpers/TwoFilesConfigurationTests.cs
ErikMogensen: I think it is best to be consistent, but like the previous issue I don't have a strong opinion about that so I changed that too.

src/ServiceBusExplorer/Helpers/MessageAndPropertiesHelper.cs
ErikMogensen: Good tip. Those lines are just moved from the MainForm.

SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.1.0-PullRequest0261.295.build.477 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/build/4.1.0-PullRequest0261.295.build.477) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/a33acfe6e4 by @ErikMogensen)

paolosalvatori:
As @SeanFeldman  said, we need to understand if saving connection strings in the user configuration file is something that could cause to customers. I have approved the PR, but it would be better to add an option to allow users to save settings in the ServiceBusExplorer.exe.config or in the user configuration file. If the user choose to use the user configuration file, we should:

 - migrate existing settings and connection strings from ServiceBusExplorer.exe.config  to the user configuration file
 - remove settings and connection strings from ServiceBusExplorer.exe.config  
 - add a link to the user configuration file in the options dialog. I don't want that users get lost searching for the user configuration file in the file system. :)

In other words, we need to be compliant with the past. I'm also a bit skeptical about using c# 7.1. This means that customers need to install the latest version available of the .NET Framework, right? I want to make sure that preserve maximum compatibility with previous versions. Thoughts?

paolosalvatori:
> having the connection strings stored as part of the roaming profile
> 
> 👍 If this will be an issue, we'll hear from users.

As I said above, let's make sure that this new behavior causes issues to existing customers. See the post I wrote on this thread.

src/ServiceBusExplorer.Tests/app.config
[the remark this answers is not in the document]
SeanFeldman: @ErikMogensen my comment wasn't about `supportedRuntime` element, but `<system.serviceModel>` and all the noise it brings when ASB client is installed.

SeanFeldman:
> As I said above, let's make sure that this new behavior causes issues to existing customers.

That we can do @paolosalvatori 😂

src/ServiceBusExplorer/Helpers/DoubleHelper.cs
SeanFeldman: If it looks like a duck, swims like a duck, and quacks like a duck, then it probably is a duck.
You could turn it into an extension, no?
```c#
double x, y;
x.NearlyEqual(y);
```

ErikMogensen:
Regarding C# version, I am pretty sure it is not a problem. I tried different combinations of the C# version setting and the Target framework setting and some combinations do not compile. It looks like the target version of the .NET Framework is solely determined by the Target framework setting (on the Application tab in project settings).

See also https://stackoverflow.com/questions/10636878/how-does-c-sharp-version-vs-version-and-net-framework-version-etc-fit-togeth

> add a link to the user configuration file in the options dialog. I don't want that users get lost searching for the user configuration file in the file system. :)

I totally agree with the above.

> migrate existing settings and connection strings from ServiceBusExplorer.exe.config to the user configuration file
> remove settings and connection strings from ServiceBusExplorer.exe.config
> 
I am not sure about the above. I know that some organizations put SBE on the central file location where everyone can access it. That way you get some settings by default and is able to override them in your user config file.

Anyone have a good idea how it should to decide if it should use the user config or the application config? Perhaps always use a setting in the application config for this.

SeanFeldman:
I don't know what people are doing with SBE in the wild, but would think personal settings should always override whatever is found in the `.exe.config`.

ErikMogensen:
Regarding the issue about choice of config file. What do you think of this? 

- Create a new Setting: ConfigFile. It has three possible values: _Application_, _User_ and _Both_.
- If the setting does not exist in any file the Application file is used. Same as before.
- If the setting exists in the User config file it overrides the value in the Application file.
- When switching from _User_ or _Both_ to _Application_ it will change the setting in both files.


The UI for the new setting is below.
image

I do not know why the dropdown list does not want to show a border. Ideas on that?

The _Open_ button will open the selected type of configuration file using the associated application. If _Both_ is selected, then both files will be opened. (It will probably be opened using Notepad which is not that great. A later version of SBE could offer a possibility to select an application for opening config files.) Another alternative is to display the paths to both of the config files in read only textboxes offering the possibility to copy it. (Perhaps also a _Copy to clipboard_ button like in the Azure Portal.)

paolosalvatori:
> Regarding the issue about choice of config file. What do you think of this?
> 
> Create a new Setting: ConfigFile. It has three possible values: Application, User and Both.
> If the setting does not exist in any file the Application file is used. Same as before.
> If the setting exists in the User config file it overrides the value in the Application file.
> When switching from User or Both to Application it will change the setting in both files.
> 
> The UI for the new setting is below.
> 
> I do not know why the dropdown list does not want to show a border. Ideas on that?
I used a trick to override this issue: I explicitly draw a rectangle around any drop-down list control in the tool using an event handler for the Paint event. Look at the mainPanel_Paint method in the OptionsForm.cs. This methods is used as event handler of the Paint event for the mainPanel control.
> The Open button will open the selected type of configuration file using the associated application. If Both is selected, then both files will be opened. (It will probably be opened using Notepad which is not that great. A later version of SBE could offer a possibility to select an application for opening config files.) Another alternative is to display the paths to both of the config files in read only textboxes offering the possibility to copy it. (Perhaps also a Copy to clipboard button like in the Azure Portal.)
To open a file using the default application defined for a given extension (e.g. config), you can use the following line of code:

System.Diagnostics.Process.Start(@"c:\user.config");

paolosalvatori:
@ErikMogensen please let me know when you have done with changes, so I can merge your edits into the develop branch. Thanks! :)

ErikMogensen:
Don't review nor merge this yet.

SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.1.0-PullRequest0261.297.build.480 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/19360801) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/f094301bf3 by @ErikMogensen)

SeanFeldman:
> Don't review nor merge this yet.

We should prefix the title with `[WIP]` and [WIP plugin](https://github.com/paolosalvatori/ServiceBusExplorer/issues/217) will block from merging.

paolosalvatori:
Ok, let me know when you have done!

SeanFeldman:
:x: [Build ServiceBusExplorer 4.1.0-PullRequest0261.301.build.482 failed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/19528991) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/2844534397 by @)

SeanFeldman:
compilation error

ErikMogensen:
Don't review this yet.

SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.1.0-PullRequest0261.302.build.483 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/19660390) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/fafe324905 by @)

SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.1.0-PullRequest0261.303.build.484 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/19660586) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/1b451903cf by @)

SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.1.0-PullRequest0261.304.build.485 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/19670021) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/c8cefe7340 by @ErikMogensen)

src/ServiceBusExplorer/Helpers/Constants.cs
ErikMogensen: Fixed

src/ServiceBusExplorer/Helpers/MessageAndPropertiesHelper.cs
ErikMogensen: That's Paolo :)

SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.1.0-PullRequest0261.305.build.486 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/19670239) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/ec82eb57a2 by @ErikMogensen)

src/ServiceBusExplorer/Helpers/XmlLinqExtensions.cs
ErikMogensen: Thanks

SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.1.0-PullRequest0261.306.build.487 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/19670352) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/b28428317e by @ErikMogensen)

ErikMogensen:
I changed the OptionForm to use a group control to differentiate from selection of the configuration file.

image

I think it would look better if all of the background was blue, but since there is a white group control covering all of the background I did not want to change it. 


I also added some information in the lower left corner on the ConnectForm:

image

There is one issue with this PR that I am aware of; when starting SBE and  user config file only has been saved as configuration and there are no namespaces in the user config file an error message about incorrect information will be logged. I do not know what's the best way to solve this. I suggest a separate issue for that.

SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.1.0-PullRequest0261.307.build.488 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/19676804) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/13337e815c by @ErikMogensen)

src/ServiceBusExplorer/Forms/ConnectForm.cs
```diff
@@ -145,6 +151,34 @@ public ConnectForm(ServiceBusHelper serviceBusHelper)
...
+                case ConfigFileUse.UserConfig:
+                    replacement = "The user config file";
+                    break;
+
+                case ConfigFileUse.BothConfig:
+                    replacement = "The application config file and the user config file";
+                    break;
+
+                default:
+                    throw new InvalidOperationException("Unexptected value");
```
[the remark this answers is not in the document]
ErikMogensen: Yes, and it is misspelled. Just made a push that has a more descriptive error message.

SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.1.0-PullRequest0261.308.build.489 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/19695337) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/3e3e98948a by @ErikMogensen)

src/ServiceBusExplorer/Forms/ConnectForm.cs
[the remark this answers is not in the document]
SeanFeldman: Looks like push didn't work.

PR #272, opened 2018-10-29
Suppressed all VS Code Analyzer warnings

### LINKED ISSUE #263: Able to run VS Code Analysis without any warnings
I think that you should be able to run the VS Code Analysis with default rules on the solution without any warnings. To handle that we probably have to suppress some warnings. In that case, should we create a separate file or put the suppressions in the code? I think it is best to put it in the code.

ErikMogensen:
To remove some of the warnings I put the DllImports in classes called NativeMethods according to the guideline at https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1060-move-p-invokes-to-nativemethods-class?view=vs-2017. 

One NativeMethods class already existed, but I also created a new NativeMethods class in the Helpers namespace since I did not want to create a dependency from the affected classes in the Controls namespace to the Helpers namespace.
 
Fixes #263.


SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.1.0-PullRequest0272.280.build.493 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/19891372) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/dbea57aa75 by @ErikMogensen)

PR #295, opened 2019-02-03
Add support for SendDisabled and ReceiveDisabled states for Queues

### LINKED ISSUE #222: Support for all four of queues suspension states
While the Azure Portal only supports the Active and Disabled state on queues, the NamespaceManager actually supports two additional states: SendDisabled and ReceiveDisabled. These two states can be incredibly useful and do not appear to be supported by the current version of ServiceBusExplorer.

I propose to tweak the current "Enable/Disable" contextual menu options, as well a all of their various counterparts to replace them with a "State" menu with all four options as a sub-menu. This isn't the best UX proposition ever, as it makes it harder/less-obvious to disable or enable a queue.

I could probably find sometime to write a PR for this in the coming week if this seems like a good implementation to you.

ahofman:
Note other entities are not impacted - only Queues support these states in ASB.

Context menu for queues now looks like so: 
image

The "Enable"/"Disable" button bottom right of screen now reads "Status" and pops the following dialog: 
image

Fixes #222


SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.2.0-PullRequest0295.83.build.514 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/22083345) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/103e565cf9 by @ahofman)

src/ServiceBusExplorer/Forms/MainForm.cs
```diff
@@ -2684,6 +2682,72 @@ private async void deleteEntity_Click(object sender, EventArgs e)
...
+                    var status = queueDescription.Status;
+                    foreach (var dropDownItem in changeStatusQueueMenuItem.DropDownItems)
+                    {
+                        var dropDownMenuItem = dropDownItem as ToolStripMenuItem;
+                        dropDownMenuItem.Checked = (EntityStatus)dropDownMenuItem.Tag == status;
+                    }
+                }
+                else
+                {
+                    throw new ApplicationException("Not supported for entity types other than queues.");
```
ErikMogensen: Great that there is check for the Tag type, but it should throw a more specific Exception. In this case I suggest NotSupportedException. I know that ApplicationException is used frequently within other places in the source code.

ErikMogensen: Great that there is check for the Tag type, but it should throw a more specific Exception. In this case I suggest NotSupportedException. I know that ApplicationException is used frequently within other places in the source code.

SeanFeldman: Agreed. `ApplicationException` is not recommended on general these days.

SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.2.0-PullRequest0295.84.build.516 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/22093037) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/20ee493aed by @ahofman)

PR #309, opened 2019-03-05
Lazy load subscription rules

### LINKED ISSUE #308: Very slow loading times for topics with a high number of subscriptions
We have a Service Bus with three topics, and each topic has approximately 1000 different subscriptions. Loading the namespace in Service Bus Explorer takes between 5 and 10 minutes with the program completely unresponsive during this time.

I've investigated the problem and it stems from the fact that for every subscription the program is doing an additional request to get the subscription's rules so that it can build the tree. In our case this leads to the following requests:

- Read list of subscriptions for **TopicA** (1 request)
- Read rules for each subscription in **TopicA** (1000 requests)
- Read list of subscriptions for **TopicB** (1 request)
- Read rules for each subscription in **TopicB** (1000 requests)
- Read list of subscriptions for **TopicC** (1 request)
- Read rules for each subscription in **TopicC** (1000 requests)

This results in 3000 additional requests. I've tinkered with the code and when I remove the rule requests the load time improves significantly, and the tree can be built in tens of seconds rather than many minutes.

Given that Service Bus supports up to 2000 subscriptions per topic I believe our use case is valid. Would it be feasible for Service Bus Explorer to implement lazy loading of the rules nodes, so that the rules request for a subscription is only sent when that particular subscription is highlighted or expanded?

sterobson:
Fixes #308 - Very slow loading times for topics with a high number of subscriptions.

Topic Subscription nodes will now load a blank "Rules" node, and the real Rules will only be read from the server when the Subscription is selected or expanded. This speeds up the loading of Topics with hundreds of Subscriptions by an order of magnitude.

This was achieved by extracting much of the GetEntities() method which builds the tree into a LazyLoadNode() method. This will still add the Subscription nodes, but then adds the Rules to a list of items still be read and processed only when a Subscription node is clicked or expanded.

### DIFF
--- a/src/ServiceBusExplorer/Forms/MainForm.cs
+++ b/src/ServiceBusExplorer/Forms/MainForm.cs
@@ -4551,7 +4551,6 @@ namespace Microsoft.Azure.ServiceBusExplorer.Forms
             if (treeNodesToLazyLoad?.Remove(node) ?? false)
             {
                 LazyLoadNode(node);
-                treeNodesToLazyLoad.Remove(node);
             }
         }
 
--- a/src/ServiceBusExplorer/Forms/MainForm.cs
+++ b/src/ServiceBusExplorer/Forms/MainForm.cs
@@ -21,6 +21,11 @@
 
 #region Using Directives
 
+using Microsoft.Azure.NotificationHubs;
+using Microsoft.Azure.ServiceBusExplorer.Controls;
+using Microsoft.Azure.ServiceBusExplorer.Enums;
+using Microsoft.Azure.ServiceBusExplorer.Helpers;
+using Microsoft.ServiceBus.Messaging;
 using System;
 using System.Collections;
 using System.Collections.Concurrent;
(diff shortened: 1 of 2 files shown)


SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.2.0-PullRequest0309.89.build.525 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/22845824) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/f4bd598124 by @)

src/ServiceBusExplorer/Forms/MainForm.cs
```diff
@@ -4606,6 +4542,122 @@ bool FilterOutException(Exception ex)
...
+        /// <summary>
+        /// If the node is in the list of nodes that still require LazyLoading then remove the node from the list and lazy load it.
+        /// </summary>
+        /// <param name="node"></param>
+        private void EnsureNodeHasBeenLazyLoaded(TreeNode node)
+        {
+            if (treeNodesToLazyLoad?.Remove(node) ?? false)
+            {
+                LazyLoadNode(node);
+                treeNodesToLazyLoad.Remove(node);
```
SeanFeldman: `if` condition will attempt to remove a given `node` from the collection `treeNodesToLazyLoad` and only if operation is successful, the body will execute. In that case, the 2nd removal is unnecessary, isn't it?

SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.2.0-PullRequest0309.90.build.526 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/22858136) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/dbff0abf62 by @)

src/ServiceBusExplorer/Forms/MainForm.cs
sterobson: Good spot, I've now removed the redundant call to `treeNodesToLazyLoad.Remove(node)`.

paolosalvatori:
Thanks @sterobson... with a delay :)

PR #329, opened 2019-03-14
Make reversion from User config file to Application config file stick

### LINKED ISSUE #319: Cannot revert Configuration file option to "Application Configuration File"
In View > Options, the option "Configuration File for Settings and Connection Strings" defaults to "Application Configuration File" (I think). At some point, I set it to "Both (User file will override)", and now I can't set it back.

I'll open View > Options, switch it to "Application Configuration File", hit Save, and hit OK. If I re-open View > Options in the same program session, it'll still be set to "Application Configuration File". But if I exit Service Bus Explorer and restart it, the log says "The files C:\Users\...\UserSettings.config and C:\ProgramData\chocolatey\...\ServiceBusExplorer.exe.Config are used for the configuration settings.", and in View > Options, it is back to "Both (user file will override)".

This is Service Bus Explorer version 4.1.113 on Windows 10. SBE is installed via chocolatey. I have uninstalled and reinstalled SBE, and it didn't make a difference.

apjanke:
Fixes #319.

When you have set the option to use the User config file or Both config file, and then switch back to Application config file and save your settings, that change will not stick through an application restart. This is because the settings, including which config file to use, get saved to the Application file and not the User config file. So the User config file still has the old "User" or "Both" setting in it, and that will get loaded again and take precedence at the next program restart.

This PR fixes it by adding special-case logic to write the ConfigFileUse setting to the User config file if we have switched from a User-using ConfigFileUse to Application ConfigFileUse during the use of the Options form.

### DIFF
--- a/src/ServiceBusExplorer/Forms/OptionForm.cs
+++ b/src/ServiceBusExplorer/Forms/OptionForm.cs
@@ -432,7 +432,7 @@ namespace Microsoft.Azure.ServiceBusExplorer.Forms
                     return BothConfigFileIndex;
 
                 default:
-                    throw new InvalidDataException("Unexpexted value passed to " +
+                    throw new InvalidDataException("Unexpected value passed to " +
                                                    nameof(GetIndexForConfigFileUseUIString));
             }
         }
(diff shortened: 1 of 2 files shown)


SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.2.0-PullRequest0329.102.build.546 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/23084100) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/125d2dc13b by @apjanke-ccs)

apjanke:
> That was quick work!

Dogfooding makes for fast improvements. ;)

paolosalvatori:
Thanks @apjanke  for the PR 🥇

Issue #327, opened 2019-03-14
Failure creating Event Hub in existing namespace: Forbidden


apjanke:
I'm having trouble creating a new Event Hub in an existing Service Bus namespace using SBE.

I have an existing ASB namespace “mrtl-rma-test-01” that contains a Queue and a Topic. I've connected successfully to it using the default RootManageSharedAccessKey.

I tried to create an Event Hub by right-clicking on the Event Hubs node in the navigation tree, selecting Create Event Hub, putting “mrtl-rma-test-eventhub-01” as the Relative URI, and clicking the Create button in the lower left.

SBE froze for a bit over 30 seconds, and then displayed some error messages in the log.

```
<12:29:56> The files C:\Users\NY510094\AppData\Roaming\Service Bus Explorer\UserSettings.config and C:\Users\NY510094\local\repos\ServiceBusExplorer\src\ServiceBusExplorer\bin\Debug\ServiceBusExplorer.exe.Config are used for the configuration settings.
<12:30:00> The application is now connected to the sb://mrtl-rma-test-01.servicebus.windows.net/ service bus namespace.
<12:30:00> MessagingFactory successfully created
<12:30:02> The queue rma-pricing-requests has been successfully retrieved.
<12:30:03> The topic rma-pricing-notifications has been successfully retrieved.
<12:37:21> Exception: The remote server returned an error: (403) Forbidden. TrackingId:823b965b-c281-40e6-9ca4-cb8e8b221c1d, SystemTracker:mrtl-rma-test-01.servicebus.windows.net:mrtl-rma-test-eventhub-01, Timestamp:3/14/2019 4:36:55 PM. Method <CreateEventHub>b__0: retry 1 of 10.
<12:37:21> Exception: The remote server returned an error: (403) Forbidden. TrackingId:4cdb9b33-5321-47a4-b8cd-8a4a5f40816e, SystemTracker:mrtl-rma-test-01.servicebus.windows.net:mrtl-rma-test-eventhub-01, Timestamp:3/14/2019 4:36:58 PM. Method <CreateEventHub>b__0: retry 2 of 10.
[...]

PR #330, opened 2019-03-14
Fix long delay when creating invalid entity

apjanke:
Partially addresses #327.

Currently, when you attempt to create an Event Hub in a Namespace that is for Queues/Topics, or vice versa, the UI will freeze for about 30 seconds, before spewing some errors to the log. This is because the creation attempt is retried 10 times, even though this isn't a transient error.

This PR improves UI responsiveness by changing `RetryHelper` to consider `QuotaExceededException` (which is what happens in this case) to be non-transient, and not retry it.

Does not fix the UI freeze when creating Relays in a non-Relay namespace, because that throws a different `AggregateException`, which I'm not comfortable special-casing to be non-transient in this PR, unless y'all give that an explicit thumbs-up.

You could go even further along this route: `RetryHelper` considers the generic `Exception` to be transient and retries it. I think maybe it should be the other way around: consider generic `Exception` to be non-transient, and only retry on special-cased exceptions that are known to be probably transient.

### DIFF
--- a/src/ServiceBusExplorer/Helpers/RetryHelper.cs
+++ b/src/ServiceBusExplorer/Helpers/RetryHelper.cs
@@ -238,6 +238,19 @@ namespace Microsoft.Azure.ServiceBusExplorer.Helpers
                                                  retryCount));
                     Thread.Sleep(retryTimeout);
                 }
+                catch (MessagingException ex)
+                {
+                    if (numRetries == 0 || (!ex.IsTransient))
+                    {
+                        throw;
+                    }
+                    writeToLog(string.Format(RetryingMethod,
+                                                 ex.Message,
+                                                 func.Method.Name,
+                                                 retryCount - numRetries + 1,
+                                                 retryCount));
+                    Thread.Sleep(retryTimeout);
+                }
                 catch (TimeoutException ex)
                 {
                     if (numRetries == 0)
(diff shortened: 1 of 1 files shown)


apjanke:
> I just saw in the source code that is is already doing the IsTransient check for MessagingException, which QuotaExceeded inherits from.

Are you sure about that? Looking at the [`QuotaExceededException` docs](https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.servicebus.quotaexceededexception?view=azure-dotnet), it looks like `QuotaExceededException` inherits from `Microsoft.Azure.ServiceBus.ServiceBusException`, which is unrelated to `MessagingException`. `ServiceBusException` has an `IsTransient` property. So instead of handling `QuotaExceededException` specially, I think we need to add a `catch` for `ServiceBusException`, and have it use the same logic as the existing `catch` for `MessagingException`.

apjanke:
Nevermind, you're right: _this_ `QuotaExceededException` (`Microsoft.ServiceBus.Messaging.QuotaExceededException`) is in fact derived from `Microsoft.ServiceBus.Messaging.MessagingException`. That [`Microsoft.Azure.ServiceBus.QuotaExceededException`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.servicebus.quotaexceededexception?view=azure-dotnet) I was looking at must be from a different SDK or a different version of the SDK.

I'll take your suggested approach instead.

ErikMogensen:
Yes, It is easy to mix the old and the new SDKs. The documentation is sometimes not clear if it is for the old, the new or both SDKs. Anyway:

- Microsoft.ServiceBus.dll: old one based on Full .NET Framework
- Microsoft.Azure.ServiceBus.dll: new one based on .NET Standard

SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.2.0-PullRequest0330.104.build.555 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/23122669) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/d6f6841a6c by @apjanke-ccs)

apjanke:
Aha! The issue is that the RetryFunc* variants (as opposed to RetryAction*) are only catching `MessagingCommunicationException`, not `MessagingException`, so `MessagingException`s, including `QuotaExceededException`, are not getting tested for `IsTransient` before doing the retry.

I've amended this commit to just add `catch (MessagingException ex)` blocks to the RetryFunc* variants. No special handling for `QuotaExceededException` now.

paolosalvatori:
Thanks @apjanke!

PR #331, opened 2019-03-15
Reduce excess precision in msg/sec rates in log messages

apjanke:
In the log messages for sending and receiving messages, SBE currently formats the “messages/second” rates using the full precision of a `double`, with about 14 digits past the decimal point.

```
     - Message Count=[1] Messages Sent/Sec=[4.20168067226891] Total Elapsed Time (ms)=[238]
```

That's a wackdoodle amount of precision for the data we're looking at here. Given the actual variance in message send/receive times, and the measurement error, I'd guess there's maybe two actual significant digits of data there.

This PR reduces the excess precision here by changing the formatting of the msgs/sec values to use just 1 decimal point. Because it'll always take at least 10-20 ms to send a message, that'll be sufficient precision, and we don't need to worry about scale-independent significant digits here.

This will reduce potential user impression of high precision in metrics data, and IMHO makes the log messages more readable.

This PR also throws in a typo fix that occurs in several “statistics”-related private variable names.

### UI change screenshots

Before:

lokfojmekbbfaoba

After:

gbjgpejffbhpndlj

### DIFF
--- a/src/ServiceBusExplorer/Controls/TestRelayControl.cs
+++ b/src/ServiceBusExplorer/Controls/TestRelayControl.cs
@@ -77,7 +77,7 @@ namespace Microsoft.Azure.ServiceBusExplorer.Controls
         private const string SendTaskCountMustBeANumber = "The Sender Task Count field must be an integer number greater than zero.";
         private const string MessageCannotBeNull = "The Message field cannot be null.";
         private const string SenderStatisticsHeader = "Sender[{0}]:";
-        private const string SenderStatisticsLine1 = " - Message Count=[{0}] Messages Sent/Sec=[{1}] Total Elapsed Time (ms)=[{2}]";
+        private const string SenderStatisticsLine1 = " - Message Count=[{0}] Messages Sent/Sec=[{1:F1}] Total Elapsed Time (ms)=[{2}]";
         private const string SenderStatisticsLine2 = " - Average Send Time (ms)=[{0}] Minimum Send Time (ms)=[{1}] Maximum Send Time (ms)=[{2}] ";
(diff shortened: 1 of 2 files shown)


SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.2.0-PullRequest0331.101.build.547 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/23089937) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/f2952082e8 by @apjanke-ccs)

apjanke:
> Suggest to keep 3 digits after the decimal point for number of mSecs

But that's not (necessarily) msec in this case. This is msgs/sec, so the unit is 1/sec. 3 digits past the decimal place only corresponds to msec if your msg/sec rate is down near 1 message per second, which I would think would be an unusual case. If your msg/sec rate is closer to 100 or 1000 msgs/sec, then msec is only 1 or 0 digits after the decimal point.

apjanke:
BTW, do you know how I can send several messages at once using SBE, so I could get log output that has multiple “Sender[n]: ...” messages in a row?

SeanFeldman:
😂 sigh, silly me.

apjanke:
> @apjanke , I don't understand your question.

So, we've got a log message that's "Sender[0]: ...". That kind of implies there might be a Sender[1] and Sender[2] and so on, if there were a big enough batch of messages to handle. I'm wondering because maybe this could be reformatted in a way that's easier to read in the case of sends or receives with multi-Sender batches.

ErikMogensen:
Ok. As far as I know there is never more than one sender[x] per row. To configure for multiple senders set the Task count setting to > 1.

PR #335, opened 2019-03-15
Fix bogus “entities exported” log message

### LINKED ISSUE #333: Bogus “entities exported to .” log message when export is canceled
If I choose “Export Queue”, “Export Entities”, or another export operation, and then Cancel the file selection dialog, this message shows up in the log:

```
<22:36:30> Selected entities have been exported to . 
```

That's wrong. Either there should be no log message, or it should say something like “Entity export canceled by user.”

Actually, it's worse than that. Here’s the full log I get when I start SBE, connect to a namespace, right-click it and choose Export Entities, and then just click Cancel in the file selection dialog.

```
<23:06:50> The files C:\Users\NY510094\AppData\Roaming\Service Bus Explorer\UserSettings.config and C:\Users\NY510094\local\repos\ServiceBusExplorer\src\ServiceBusExplorer\bin\Debug\ServiceBusExplorer.exe.Config are used for the configuration settings.
<23:06:55> The application is now connected to the sb://mrtl-rma-test-01.servicebus.windows.net/ service bus namespace.
<23:06:55> MessagingFactory successfully created.
<23:06:57> The queue rma-pricing-requests has been successfully retrieved.
<23:06:58> The topic rma-pricing-notifications has been successfully retrieved.
<23:07:04> The queue rma-pricing-requests has been successfully exported.
<23:07:04> The topic rma-pricing-notifications has been successfully exported.
<23:07:07> Selected entities have been exported to . 
```

Those two “has been successfully exported” entries probably shouldn’t be there, because nothing has been exported to an actual file yet. The problem is that those log messages are generated when the entity objects are converted to in-memory XML in Strings by `ImportExportHelper` in `SerializeEntity`. I don’t think that’s a significant enough event to log.

apjanke:
Fixes #333.

Currently, if you right-click a node in the navigation tree, select “Export <whatever>”, and then Cancel the file selection dialog, the export will abort, but you will still get log messages about “... has been successfully exported” and “Selected entities have been exported to .”.

The first is because the log messages happen when in-memory serialization from objects to XML strings happens (not when a file is written). The second is because the export code doesn't check whether the Save dialog was canceled.

This PR fixes this by:
* Adding a check for whether the Save dialog was cancelled, and suppressing the log message upon cancellation.
* Removing all the “successfully exported” log messages for XML serialization.

I realize that the latter is making a big change, but I think it's the right thing to do: that XML serialization method doesn't know whether it's happening in-memory or going to an external data store. Only the caller knows if an export to an external data store really happens, and that's the only thing that the user cares about. And I don't think the user cares about getting a separate log notification for each type of thing that was converted to XML.

###  Screenshots

This is what happens if I:
1. Lauch SBE
2. Connect to an ASB namespace
3. Right-click the namespace and choose Export Entitites
4. Cancel the file selection dialog

Before:

kjiijkclamcepane

After:

bflhbabpikhfdckb

### DIFF
--- a/src/ServiceBusExplorer/Forms/MainForm.cs
+++ b/src/ServiceBusExplorer/Forms/MainForm.cs
@@ -5323,7 +5323,11 @@ namespace Microsoft.Azure.ServiceBusExplorer.Forms
             var path = entityType == null ?
                        CreateFileName(string.Format(EntitiesFileNameFormat, serviceBusHelper.Namespace, entityName)) :
                        CreateFileName(string.Format(EntityFileNameFormat, serviceBusHelper.Namespace, entityName, entityType));
-            WriteToLog(string.Format(EntitiesExported, SaveEntityToFile(xml, path)));
+            var savedFile = SaveEntityToFile(xml, path);
+            if (savedFile != null)
+            {
+                WriteToLog(string.Format(EntitiesExported, savedFile));
+            }
         }
 
         private void copyStringToClipboard(string str)
--- a/src/ServiceBusExplorer/Helpers/ImportExportHelper.cs
+++ b/src/ServiceBusExplorer/Helpers/ImportExportHelper.cs
@@ -642,68 +642,6 @@ namespace Microsoft.Azure.ServiceBusExplorer.Helpers
                 }
             }
             xmlWriter.WriteEndElement();
-            switch (typeName)
-            {
-                case QueueDescriptionClass:
-                    var queueDescription = entity as QueueDescription;
-                    if (queueDescription != null)
-                    {
-                        MainForm.StaticWriteToLog(string.Format(QueueExported, queueDescription.Path));
-                    }
-                    break;
-                case TopicDescriptionClass:
-                    var topicDescription = entity as TopicDescription;
-                    if (topicDescription != null)
-                    {
-                        MainForm.StaticWriteToLog(string.Format(TopicExported, topicDescription.Path));
-                    }
-                    break;
-                case RelayDescriptionClass:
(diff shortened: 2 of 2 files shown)


SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.2.0-PullRequest0335.102.build.550 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/23091542) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/92d3956ffc by @apjanke-ccs)

SeanFeldman:
I must be blind, but this PR fixies a bunch of typos, but not the change it defines in the issue description. Am I missing it?

apjanke:
Are you looking at all three commits?

Here's the change that does the actual fix; it's in the third commit: https://github.com/paolosalvatori/ServiceBusExplorer/pull/335/commits/2f1415d049c9dd0f28fbd00f169cb8941c8c27c6#diff-c6c670cf209d398a1828a3de03a47565R5327.

It takes this unconditional `WriteToLog` call:

```
            WriteToLog(string.Format(EntitiesExported, SaveEntityToFile(xml, path)));
```

and conditionalizes it on whether `SaveEntityToFile` actually did something:

```
            var savedFile = SaveEntityToFile(xml, path);
            if (savedFile != null)
            {
                WriteToLog(string.Format(EntitiesExported, savedFile));
            }
```

Then there's this big deletion: https://github.com/paolosalvatori/ServiceBusExplorer/pull/335/commits/c63bd5d035e4725d8137adaa626c34e26211fbe6. It removes the log output for when XML serialization is just done in-memory.

PR #379, opened 2019-09-17
WIP: Split

ErikMogensen:
Split the Service Bus Explorer into some assemblies. 
Use the new .NET Standard based libraries for Event Hubs, Relay and Service Bus.
Moved some classes into the new assemblies.


ErikMogensen:
Replaced this PR with a later one.

PR #385, opened 2019-10-15
Prevent using an entity-specific connection string

### LINKED ISSUE #376: Do not allow entity-specific connection strings
Feature request #375 has highlighted that today the tool is handling entity-specific connections string in a non-graceful way. The tool should prevent entity-specific connection strings from being used.

A suggestion made in https://github.com/paolosalvatori/ServiceBusExplorer/issues/375#issuecomment-530040760 could be an option.

lenardg:
This PR aims to fix #376 

Having run into this issue myself a few times, I decided to fix it.

While in the Save case I opted to reuse the logic provided by the connection string builder there was no such logic in the Ok button handler - the connection string is processed outside the dialog. Because it made sense to show the warning before the dialog is closed and because I did not want to add the builder there, I decided to add a simple regex verification in this case.

### DIFF
--- a/src/ServiceBusExplorer/Helpers/ConnectionStringHelper.cs
+++ b/src/ServiceBusExplorer/Helpers/ConnectionStringHelper.cs
@@ -1,0 +1,19 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.ServiceBusExplorer.Helpers
+{
+    public static class ConnectionStringHelper
+    {
+        private readonly static Regex EntityPathRegex = new Regex(@"(?:^|[^\w])(EntityPath=[^;]+)");
+        
+        public static bool IsEntitySpecific(string connectionString)
+        {
+            return EntityPathRegex.IsMatch(connectionString);
+        }
+    }
+}
--- a/src/ServiceBusExplorer/Helpers/ConnectionStringHelper.cs
+++ b/src/ServiceBusExplorer/Helpers/ConnectionStringHelper.cs
@@ -1,16 +1,15 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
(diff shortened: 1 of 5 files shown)


SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.2.0-PullRequest0385.120.build.595 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/28118223) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/c022ed26f2 by @lenardg)

src/ServiceBusExplorer/Helpers/ConnectionStringHelper.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.ServiceBusExplorer.Helpers
+{
+    public static class ConnectionStringHelper
```
SeanFeldman: Given it would not be very difficult to unit-test this logic, how about adding a test to verify `ConnectionStringHelper` works as expected? There's already a unit test project you could use.
Thank you.

SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.2.0-PullRequest0385.122.build.596 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/28140484) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/7ca7b116f0 by @lenardg)

src/ServiceBusExplorer/Helpers/ConnectionStringHelper.cs
lenardg: Sure thing. I added the unit tests for the new method. 

I also pushed an additional commit to clean up  `using` statements in the new helper class file.

SeanFeldman:
@paolosalvatori please review/merge

PR #401, opened 2019-12-28
Enhance forwards UI experience

Greybird:
Hello,

This PR is a proposal to better handle structured hierarchies of queues and topics when dealing with forwards.
* it exposes the **Path** relative to the namespace for **ForwardTo** and **ForwardDeadLetteredMessagesTo**
* it also ensures that when opening the queue/topic selection box for these auto-forward configuration, if an existing queue or topic is already configured, it will be focused.

I had to fix a bug in the auto focus code of the selection form to allow searching below the second level, which might contain only the part before the last slash in the queue/topic path.

Also note that this last feature will require network access to get the queue or topic from the configured Path. However, as this access will be done only on the opening of the form, I consider this is acceptable compared to the enhanced experience.

# Screenshots
UI showing that ForwardTo displays the whole Path instead of just the name:
image

Selection UI just after clicking on the ... button. Currently configured queue is displayed:
image

### DIFF
--- a/src/Common/Helpers/ServiceBusHelper.cs
+++ b/src/Common/Helpers/ServiceBusHelper.cs
@@ -2289,7 +2289,7 @@ namespace Microsoft.Azure.ServiceBusExplorer
         /// Gets the uri of the deadletter queue for a given queue.
         /// </summary>
         /// <param name="queuePath">The name of a queue.</param>
-        /// <returns>he absolute uri of the deadletter queue.</returns>
+        /// <returns>the absolute uri of the deadletter queue.</returns>
         public Uri GetQueueDeadLetterQueueUri(string queuePath)
         {
             if (IsCloudNamespace)
--- a/src/Common/Helpers/ServiceBusHelper.cs
+++ b/src/Common/Helpers/ServiceBusHelper.cs
@@ -5594,14 +5594,17 @@ namespace Microsoft.Azure.ServiceBusExplorer
             if (Uri.IsWellFormedUriString(address, UriKind.Absolute))
             {
                 var uri = new Uri(address, UriKind.Absolute);
-                var uriRelativeToNamespace = NamespaceUri.MakeRelativeUri(uri);
-                return uriRelativeToNamespace.ToString();
+                if (NamespaceUri.IsBaseOf(uri))
+                {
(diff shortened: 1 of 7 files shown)


SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.2.0-PullRequest0401.129.build.614 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/29789278) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/86c8576f91 by @)

src/Common/Helpers/ServiceBusHelper.cs
```diff
@@ -5589,6 +5589,22 @@ public string GetHostWithoutNamespace()
             return host.Substring(index);
         }
 
+        public string GetAddressRelativeToNamespace(string address)
```
SeanFeldman: This code would be a good candidate for testing along with other helpers (https://github.com/paolosalvatori/ServiceBusExplorer/tree/develop/src/ServiceBusExplorer.Tests/Helpers).

SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.2.0-PullRequest0401.130.build.616 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/29807881) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/a5c0760fcf by @)

Greybird:
Hi,

Thanks for your feedback!

I wrote some tests, and doing this I discovered some small issues with the previous version of the code. I did some little changes to adapt to these.

Note that most of the testing is related to cases where the Forward uri are relative and/or not under the servicebus current namespace, which is not supposed to happen under current ServiceBus version.
However, this makes the code more resilient to a future change, even if in this case, the data provided will be less accurate (last segment or the path instead of the full path).

Regards

SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.2.0-PullRequest0401.131.build.617 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/29812006) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/7fc326efef by @)

SeanFeldman:
Tests look great 👍

paolosalvatori:
Thanks @Greybird

PR #418, opened 2020-01-25
Change Options form buttons

### LINKED ISSUE #270: Change behavior and text for some of the OptionsForm buttons
To persist the changes you make in the Options dialog you have to click the Save button. That is not like how it works in other programs and it is probably confusing since settings won't be kept between sessions if you don't click Save.

The OK button should save the settings to disk and then do the same things as before.

Instead the Save button should moved up to the right corner to be besides the new Open button. That way you can see your changes in the file(s) if you click Save and then Open.

ErikMogensen:
This fixes #270. 

The buttons now look like this:
image

The OK button now saves the values to file. The Reset button sets the parameters to their defaults.


SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.2.0-ChangeOptionsFormButtons.53.build.631 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/30352087) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/4f802e43bd by @ErikMogensen)

PR #426, opened 2020-03-01
Fix for issue 425. Displays the original json if it is malformed.

### LINKED ISSUE #425: Bad json (duplicate keys) are ignored silently
Malformed JSON is consumed by SBE and because of a change in JSON.net, silently consumes the bad json and shows the converted JSON

```
{
      'FirstName':'John',
      'FirstName':'Rambo'
}
```

Causes SBE to display the json as 

```
{
     'FirstName':'Rambo'
}
```
Expected: SBE should show the actual JSON in the body, as otherwise, one would wrongly assume the sender is sending valid data. (In my case, we were sending this payload to a java consumer and it was throwing an error, and when we looked at the message in SBE, we could not figure out what was going on).

There is a version out there that behaves correctly, and displays the malformed JSON in SBE. But the latest 4.1.112 release doesnt.

See:

1. [Json.net no longer throws in case of duplicate](https://stackoverflow.com/questions/41942171/json-net-no-longer-throws-in-case-of-duplicate) and
2. [The Json parser should throw an exception when encountering duplicate properties. #437
](https://github.com/JamesNK/Newtonsoft.Json/issues/931#issuecomment-224605585)

I believe I have a fix ready for review and will be submitting it shortly

rajrao:
Fixes #425

### DIFF
--- a/src/ServiceBusExplorer.Tests/Helpers/JsonSerializerHelperTest.cs
+++ b/src/ServiceBusExplorer.Tests/Helpers/JsonSerializerHelperTest.cs
@@ -70,7 +70,17 @@ namespace ServiceBusExplorer.Tests.Helpers
 }";
             var indented = JsonSerializerHelper.Indent(json);
 
-            Assert.AreEqual(indented, expectedResult);
+            Assert.AreEqual(expectedResult, indented);
+        }
+
+        [Test]
+        public void IndentJson_ValueIsMalformedJson_ReturnsOriginalJson()
+        {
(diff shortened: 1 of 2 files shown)


SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.2.0-Issue425-MalformedJsonIsSilentlyIgnored.55.build.641 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/31156837) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/1bf0a15056 by @rajrao)

PR #429, opened 2020-03-12
Allow updating saved service bus namespaces

Greybird:
This PR aims to allow updating existing service bus namespaces configuration.

It also:
- gets rid of some unused code (please tell me if I am mistaken)
- ensures Namespace field is read-only and synchronized with the URI to ensure the ui stays consistent upon updates

# Screenshots

You can see synchronization in action, and also the `Save` button for user created namespaces which is now visible.
connectform


SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.2.0-update-namespaces.59.build.645 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/31419392) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/3bdb349723 by @)

ErikMogensen:
I don't think this will work with Event Hubs running on Azure Stack. To handle that we would need to be able to change the URL. Please make the URL field editable.

Greybird:
> I don't think this will work with Event Hubs running on Azure Stack. To handle that we would need to be able to change the URL. Please make the URL field editable.
Hello!
Thanks for the review and feedback.
However, I'm not sure to get your comment as it is actually the namespace field which I made readonly, as it is always deduced from the uri, and has no effect on update.
However, if I missed something, could you please clarify ?

Thanks

ErikMogensen:
Ah, my bad. I misunderstood. All good.

paolosalvatori:
Same as before, it looks like this PR is in contrast with the new tabs added by Erik, right? I'll close this PR for now. If I'm wrong, please reopen it!

paolosalvatori:
Thanks @Greybird :)

PR #437, opened 2020-03-24
Added UI flow to reduce the likelihood of accidental mass deletions

logiclrd:
I was recently working with our production environment with Service Bus Explorer, and an accidental Delete keystroke was sent to the tree view with the root node selected, and I realized with alarm that I was one errant mouse or key action away from completely annihilating every single Service Bus queue, topic, subscription for our production environment.

This PR adds a new feature so that if a deletion has a broad scope, the delete form prompts the user to explicitly type in what it is they want to delete. If they don't actually type the words "Absolutely Everything", for instance, then Service Bus Explorer will not in fact delete absolutely everything.

When I was first writing this, I imagined actually putting the name of the item to be deleted into the input box -- this could still be done, but I couldn't think of obvious ways to a) keep the text short, and b) deal with tree nodes like "Queues" and "Topics". I am open to suggestions in this area -- as of right now, I gave up on trying to inject customized names and instead just use strings like "Absolutely Everything" and "All Topics In Path".

I have tried to follow the code style I saw elsewhere as best as possible. Let me know if any aspect of this doesn't pass muster. :-)

### DIFF
--- a/src/ServiceBusExplorer/Forms/DeleteForm.cs
+++ b/src/ServiceBusExplorer/Forms/DeleteForm.cs
@@ -39,6 +39,10 @@ namespace ServiceBusExplorer.Forms
         private const string Unknown = "Unknown";
         #endregion
 
+        #region Private Fields
+        private bool useAccidentalDeletionPreventionCheck = false;
+        #endregion
+
         #region Public Constructor
         public DeleteForm(string message)
         {
@@ -57,11 +61,36 @@ namespace ServiceBusExplorer.Forms
         }
         #endregion
 
+        #region Public Methods
+        public void ShowAccidentalDeletionPreventionCheck(string entityName)
+        {
+            useAccidentalDeletionPreventionCheck = true;
+
+            accidentalDeletionPreventionCheckControl.EntityName = entityName;
+            accidentalDeletionPreventionCheckControl.Visible = true;
+
+            accidentalDeletionPreventionCheckControl.Top = mainPanel.Bottom;
+            buttonsPanel.Top = accidentalDeletionPreventionCheckControl.Bottom;
+
+            ClientSize = new Size(ClientSize.Width, buttonsPanel.Bottom);
+        }
+        #endregion
+
         #region Event Handlers
         private void btnOk_Click(object sender, EventArgs e)
         {
-            DialogResult = DialogResult.OK;
-            Close();
+            bool acceptForm = false;
+
+            if (useAccidentalDeletionPreventionCheck == false)
+                acceptForm = true;
+            else
+                acceptForm = accidentalDeletionPreventionCheckControl.CheckAcceptanceAndNotifyUser();
+
+            if (acceptForm)
+            {
+                DialogResult = DialogResult.OK;
+                Close();
+            }
         }
 
         private void btnCancel_Click(object sender, EventArgs e)
--- a/src/ServiceBusExplorer/Forms/DeleteForm.cs
+++ b/src/ServiceBusExplorer/Forms/DeleteForm.cs
@@ -24,6 +24,7 @@
 using System;
 using System.Drawing;
 using System.Windows.Forms;
+using ServiceBusExplorer.Helpers;
 
 #endregion
 
@@ -40,6 +41,7 @@ namespace ServiceBusExplorer.Forms
         #endregion
 
         #region Private Fields
+        private TwoFilesConfiguration configuration;
         private bool useAccidentalDeletionPreventionCheck = false;
         #endregion
 
@@ -62,17 +64,28 @@ namespace ServiceBusExplorer.Forms
         #endregion
 
         #region Public Methods
-        public void ShowAccidentalDeletionPreventionCheck(string entityName)
+        public void ShowAccidentalDeletionPreventionCheck(TwoFilesConfiguration configuration, string deletionScopePromptText)
         {
-            useAccidentalDeletionPreventionCheck = true;
+            this.configuration = configuration;
 
-            accidentalDeletionPreventionCheckControl.EntityName = entityName;
-            accidentalDeletionPreventionCheckControl.Visible = true;
+            bool disableAccidentalDeletionPrevention = configuration.GetBoolValue(
+                ConfigurationParameters.DisableAccidentalDeletionPrevention,
+                defaultValue: false);
 
-            accidentalDeletionPreventionCheckControl.Top = mainPanel.Bottom;
-            buttonsPanel.Top = accidentalDeletionPreventionCheckControl.Bottom;
+            if (!disableAccidentalDeletionPrevention)
+            {
+                useAccidentalDeletionPreventionCheck = true;
 
-            ClientSize = new Size(ClientSize.Width, buttonsPanel.Bottom);
+                accidentalDeletionPreventionCheckControl.DeletionScopePromptText = deletionScopePromptText;
+                accidentalDeletionPreventionCheckControl.Visible = true;
+
+                accidentalDeletionPreventionCheckControl.Top = mainPanel.Bottom;
+                buttonsPanel.Top = accidentalDeletionPreventionCheckControl.Bottom;
+
+                accidentalDeletionPreventionCheckControl.Width = ClientSize.Width;
+
+                ClientSize = new Size(ClientSize.Width, buttonsPanel.Bottom);
+            }
         }
         #endregion
 
@@ -84,7 +97,18 @@ namespace ServiceBusExplorer.Forms
             if (useAccidentalDeletionPreventionCheck == false)
                 acceptForm = true;
             else
+            {
                 acceptForm = accidentalDeletionPreventionCheckControl.CheckAcceptanceAndNotifyUser();
+
+                if (accidentalDeletionPreventionCheckControl.DisableFurtherChecks)
+                {
+                    configuration.SetValue(
+                        ConfigurationParameters.DisableAccidentalDeletionPrevention,
+                        value: true);
+
+                    configuration.Save();
+                }
+            }
 
             if (acceptForm)
             {
--- a/src/Common/Helpers/ConfigurationHelper.cs
+++ b/src/Common/Helpers/ConfigurationHelper.cs
@@ -199,6 +199,9 @@ namespace ServiceBusExplorer.Helpers
             resultProperties.EncodingType = configuration.GetEnumValue
                 (ConfigurationParameters.Encoding, currentSettings.EncodingType, writeToLog);
 
+            resultProperties.DisableAccidentalDeletionPrevention = configuration.GetBoolValue
+                (ConfigurationParameters.DisableAccidentalDeletionPrevention, currentSettings.DisableAccidentalDeletionPrevention, writeToLog);
+
             resultProperties.ProxyOverrideDefault = configuration.GetBoolValue(ConfigurationParameters.ProxyOverrideDefault, currentSettings.ProxyOverrideDefault, writeToLog);
             resultProperties.ProxyUseDefaultCredentials = configuration.GetBoolValue(ConfigurationParameters.ProxyUseDefaultCredentials, currentSettings.ProxyUseDefaultCredentials, writeToLog);
             resultProperties.ProxyBypassOnLocal = configuration.GetBoolValue(ConfigurationParameters.ProxyBypassOnLocal, currentSettings.ProxyBypassOnLocal, writeToLog);
--- a/src/Common/Helpers/ConfigurationParameters.cs
+++ b/src/Common/Helpers/ConfigurationParameters.cs
@@ -57,6 +57,7 @@ namespace ServiceBusExplorer.Helpers
         public const string Encoding = "encoding";
         public const string SelectedEntitiesParameter = "selectedEntities";
         public const string MicrosoftServiceBusConnectionString = "Microsoft.ServiceBus.ConnectionString";
+        public const string DisableAccidentalDeletionPrevention = "disableAccidentalDeletionPrevention";
 
         public const string ProxyOverrideDefault = "Proxy.OverrideDefault";
         public const string ProxyAddress = "Proxy.Address";
--- a/src/Common/Helpers/MainSettings.cs
+++ b/src/Common/Helpers/MainSettings.cs
@@ -65,6 +65,8 @@ namespace ServiceBusExplorer.Helpers
         public bool UseAmqpWebSockets { get; set; }
         public Enums.EncodingType EncodingType { get; set; }
 
+        public bool DisableAccidentalDeletionPrevention { get; set; }
+
         public bool ProxyOverrideDefault { get; set; }
         public string ProxyAddress { get; set; }
         public string ProxyBypassList { get; set; }
@@ -271,6 +273,9 @@ namespace ServiceBusExplorer.Helpers
 
                 case ConfigurationParameters.UseAmqpWebSockets:
                     return UseAmqpWebSockets;
+
+                case ConfigurationParameters.DisableAccidentalDeletionPrevention:
+                    return DisableAccidentalDeletionPrevention;
 
                 case ConfigurationParameters.Encoding:
                     return EncodingType;
--- a/src/ServiceBusExplorer/Controls/AccidentalDeletionPreventionCheckControl.Designer.cs
+++ b/src/ServiceBusExplorer/Controls/AccidentalDeletionPreventionCheckControl.Designer.cs
@@ -30,9 +30,10 @@
         {
 			this.pnlFlowLayout = new System.Windows.Forms.FlowLayoutPanel();
 			this.lblMessage = new System.Windows.Forms.Label();
-			this.lblEntityName = new System.Windows.Forms.Label();
-			this.txtEntityNameCheck = new System.Windows.Forms.TextBox();
+			this.lblDeletionScopePromptText = new System.Windows.Forms.Label();
+			this.txtDeletionScopePromptCheck = new System.Windows.Forms.TextBox();
 			this.lblExplanation = new System.Windows.Forms.Label();
+			this.chkDisableAccidentalDeletionPrevention = new System.Windows.Forms.CheckBox();
 			this.pnlFlowLayout.SuspendLayout();
 			this.SuspendLayout();
 			// 
@@ -42,15 +43,15 @@
             | System.Windows.Forms.AnchorStyles.Right)));
 			this.pnlFlowLayout.AutoSize = true;
 			this.pnlFlowLayout.Controls.Add(this.lblMessage);
-			this.pnlFlowLayout.Controls.Add(this.lblEntityName);
-			this.pnlFlowLayout.Controls.Add(this.txtEntityNameCheck);
+			this.pnlFlowLayout.Controls.Add(this.lblDeletionScopePromptText);
+			this.pnlFlowLayout.Controls.Add(this.txtDeletionScopePromptCheck);
 			this.pnlFlowLayout.Controls.Add(this.lblExplanation);
+			this.pnlFlowLayout.Controls.Add(this.chkDisableAccidentalDeletionPrevention);
 			this.pnlFlowLayout.FlowDirection = System.Windows.Forms.FlowDirection.TopDown;
 			this.pnlFlowLayout.Location = new System.Drawing.Point(0, 0);
 			this.pnlFlowLayout.Name = "pnlFlowLayout";
-			this.pnlFlowLayout.Size = new System.Drawing.Size(484, 100);
+			this.pnlFlowLayout.Size = new System.Drawing.Size(484, 141);
 			this.pnlFlowLayout.TabIndex = 3;
-			this.pnlFlowLayout.SizeChanged += new System.EventHandler(this.pnlFlowLayout_SizeChanged);
 			// 
 			// lblMessage
 			// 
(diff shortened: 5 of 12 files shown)


SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.2.0-PreventAccidentalMassDeletions.63.build.686 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/31686809) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/9cc42c50ed by @logiclrd)

src/ServiceBusExplorer/Forms/DeleteForm.cs
```diff
@@ -57,11 +61,36 @@ public DeleteForm(string entityName, string entityType)
         }
         #endregion
 
+        #region Public Methods
+        public void ShowAccidentalDeletionPreventionCheck(string entityName)
```
SeanFeldman: The name `entityName` is not reflecting what's actually passed in. Which is a phrase to indicate all entities.

SeanFeldman:
One more thing - to help reviewers with UI changes, it would be good to add screenshots to visualize what the PR is accomplishing. Thank you 🙂

logiclrd:
> What I would suggest is to modify this PR to allow the following:
> 
> Make the feature configurable
> The feature to be turned off by default with a warning logged upon startup
> No hardcoded expected inputs. Those should be also configurable. (e.g. I'd use all ass the go-ahead keyword for anything).

My initial response to this is that it feels like overthinking it. Are there that many people out there who _regularly_ delete _every single topic_, ,_every single queue_, or even _the entirety of the Service Bus namespace_? This new feature doesn't trigger when you delete a single topic or queue, only when you are requesting mass destruction.

If you really think that that scenario comes up often enough that prompting people would be a significant disruption to their work, I can make further changes. My gut would be to enable the feature by default, but include a checkbox "Do not ask again" in the control.

I also feel that making the inputs configurable is over-engineering, and the whole point of making the user type in a different value for different kind of deletions is to make sure the user realizes what exactly it is they're deleting -- and using `all` for all possible values defeats that purpose. A section in the configuration where different values for different kinds of deletions can be specified would resolve this, but again, it feels like overengineering to anticipate this need and implement it in the first pass.

src/ServiceBusExplorer/Forms/DeleteForm.cs
logiclrd: That's fair. When I was initially writing this, I was anticipating having the call site indicate the actual name of the entity being deleted, but when I ran into tree nodes like `Topics` and `Queues`, I couldn't think of a way to combine that with the actual namespace name that wasn't terribly verbose and far too much text to type, so the call sites ended up collapsing down to the strings in the actual commit, like `All Topics` or `Absolutely Everything`. I think I'll change this to `deletionScopePromptText`.

SeanFeldman: `deletionScopePromptText` 👍

SeanFeldman:
> My initial response to this is that it feels like overthinking it. Are there that many people out there who regularly delete every single topic, every single queue, or even the entirety of the Service Bus namespace? This new feature doesn't trigger when you delete a single topic or queue, only when you are requesting mass destruction.

My comment is based on personal and other experiences. When working in the lower environments and need to purge either all entities of a given type or the entire namespace, delete everything is a common operation. Therefore it is not overthinking.

> If you really think that that scenario comes up often enough that prompting people would be a significant disruption to their work, I can make further changes. My gut would be to enable the feature by default, but include a checkbox "Do not ask again" in the control.

I don't think, I know 😉
Keep in mind you're adding a feature to a tool that exists for quite a while and overthrowing others' workflow w/o consideration should not happen. A checkbox to disable the feature would be less intrusive, indeed.

> I also feel that making the inputs configurable is over-engineering, and the whole point of making the user type in a different value for different kind of deletions is to make sure the user realizes what exactly it is they're deleting -- and using all for all possible values defeats that purpose. A section in the configuration where different values for different kinds of deletions can be specified would resolve this, but again, it feels like overengineering to anticipate this need and implement it in the first pass.

Do not hurt others. I am refusing to type in `Absolutely Everything`. If you wish, make it simpler - type in the node name. Look what Azure portal is doing - want to remove a resource? Type its name for the confirmation. Simple and acceptable. Thank you.

logiclrd:
I don't deny what you say about your workflow and having a regular need to delete everything, though it would take quite a bit to convince me that a significant proportion of the users of the tool regularly mass delete everything. :-)

Regarding the text to be entered, I actually changed it to `Absolutely Everything` because that is _shorter_ (by about half) than simply using the root node name in our situation. As for the other nodes, `All Topic` instead of `Topics` and `All Queues` instead of `Queues` seems to much more accurately capture what it is that is being deleted.

Ultimately, I think that if you have the option of turning the feature off entirely, then when the feature is on, having it require a meaningful input is reasonable. Do you disagree?

SeanFeldman:
What do you think about adding the option discussed above first and then review it where text changes are required or not?
If one can easily dismiss the warning and keep it that way the text won't matter much.

logiclrd:
Sounds like a plan :-) Commit coming up.

SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.2.0-PreventAccidentalMassDeletions.64.build.687 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/31782036) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/35eb078c23 by @logiclrd)

ErikMogensen:
@logiclrd , thanks for implementing this feature. I have heard of a couple mistakes people have made that this feature would prevent so I have had the intention of creating an issue for it. 

I agree with Sean's idea that you should type the name of the resource you are going to affect. That makes you see the name of it once more before it is too late and as mentioned, that is the way the Azure Portal works. To make it easier to enter the text I suggest that the text stating the name of the resource should be selectable so that you could copy the resource name and then paste it.

paolosalvatori:
Hi @logiclrd could you please include an explanation of the exact sequence of actions and screenshots? Take into account that the UI needs to be compliant with the dialogs shown when deleting a single entity. :) Frankly speaking in 8 years I never heard anyone needing to use my tool to delete all the entities from a namespace or all the entities of a certain kind. I think that in that case you should rather create a PowerShell script or an Azure CLI + Bash script :) Said that, okease provide the information I asked you. Thanks!

logiclrd:
@ErikMogensen I initially had written it so that you'd type the name, but:

- At the root, the text is exceptionally long. If I just grab the root node's text, for instance, in my debugging to verify the prompt I would have had to type `sb://iqdevelopment.servicebus.windows.net/`. This could perhaps be mitigated by only including the first component of the hostname, though naively, this would require the code to assume the format of the root node's text.

- At other levels, nodes have text like `Topics` and `Queues`. Just using the node text doesn't seem adequate. Should this then somehow be merged with the value parsed out at the root? Would you want text like, `All Topics In iqdevelopment`?

After running into these, it seemed that making the code use the actual text would increase its complexity, and I chose to keep the code simple at the expense of customized checks. If you feel that the benefits of customized checks outweigh the added complexity, I can implement a strategy to generate text based on the context -- but in most cases an actual decision will have to be made about what text to use, because the node itself is just a single word.

logiclrd:
@paolosalvatori 

With this new feature, deletions that remove entities _in bulk_ include an additional prompt right within the delete form. For instance, if you select the root node and press Delete, you get a prompt which looks like this:

image

The text to be entered varies based on which node is selected. There is also a discussion on this PR about altering what text the user is prompted to enter.

If the textbox does not contain the exact same text as the prompt, then the OK button will not permit the deletion to proceed:

image

The user can cancel the deletion and no action will be taken.

If the user has entered the exact text (not case sensitive), then the deletion proceeds as it would have before. In addition, if the checkbox ("without this check next time") is checked off, then a configuration setting is set which prevents the prompt from being shown in future deletions. This setting can be controlled from the Options form:

image

This new functionality is active for the following kinds of deletions (all in `deleteEntity_Click` in `MainForm.cs`):

* Deleting absolutely everything (root node).
* Deleting all queues.
* Deleting all topics.
* Deleting all relays.
* Deleting all event hubs.
* Deleting all consumer groups.
* Deleting all notification hubs.
* Deleting all queues in path (URL segment node).
* Deleting all topics in path (URL segment node).
* Deleting all relays in path (URL segment node).
* Deleting all rules for an entity (rules node).

It is not active for the following kinds of deletions:

* Deleting an authorization rule (`HandleEventHubControl.cs`, `HandleNotificationHubControl.cs`, `HandleQueueControl.cs`, `HandleRelayControl.cs`, `HandleTopicControl.cs`).
* Updating registrations (`HandleNotificationHubControl.cs`).
* Deleting registrations (`HandleNotificationHubControl.cs`).
* Deleting and purging queue messages and queue deadletter messages (`HandleQueueControl.cs`).
* Deleting and purging subscription messages and subscription deadletter messages (`HandleSubscriptionControl.cs`).
* Deleting a saved service bus namespace (`ConnectForm.cs`).
* Deleting all subscriptions to a topic (`MainForm.cs`).
* Deleting all consumer groups in an event hub (`MainForm.cs`).

SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.2.0-PreventAccidentalMassDeletions.66.build.691 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/31825684) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/d3eb1e0530 by @logiclrd)

ErikMogensen:
This is great but I still want to type the name of the namespace/entity. When deleting a Service Bus Namespace from the Azure Portal you have to type just the name of the namespace, not the whole URL. Below is a screenshot when I am about to remove a Service Bus Namespace called RemoveTest.
image

logiclrd:
Alright :-) What should be entered when you select the `Topics` node, in order to convey that you are deleting _all topics_, but _just topics_ of the namespace? `All %s Topics` ?

logiclrd:
And, can the root node ever have a form other than `sb://______.servicebus.windows.net/` ?

logiclrd:
Here's a proposal :-)

SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.2.0-PreventAccidentalMassDeletions.67.build.692 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/31833222) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/fc4cf7c7c0 by @logiclrd)

SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.2.0-PreventAccidentalMassDeletions.67.build.693 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/31833261) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/d5ccc7174b by @logiclrd)

logiclrd:
(force push: https://github.com/logiclrd/ServiceBusExplorer/compare/04707ff22a..f0a9ac69c)

SeanFeldman:
@logiclrd a minor correction to the checkbox. Rather than "Delete multiple items without this check next time", I would suggest to have a "Delete multiple items and **disable this check**". The difference would be the ability to dismiss this validation from the _current_ operation and on. Thanks.

logiclrd:
The way I designed it was actually intentional. I think you should be forced to type the text in that scenario. Why? Because otherwise accidentally clicking the checkbox bypasses the safety.

SeanFeldman:
> The way I designed it was actually intentional. I think you should be forced to type the text in that scenario. Why? Because otherwise accidentally clicking the checkbox bypasses the safety.

@logiclrd, I'm sorry, but I cannot approve that decision. Personal fact - I use the tool a lot in development, not in production. Forcing to type that phrase for non-production namespaces is just unnecessary friction.

logiclrd:
Okay, but you can go into the options and check off "Disable Accidental Deletion Prevention". If you do this before trying any deletions, you will _never_ get prompted, not even just once the first time.

SeanFeldman:
> Okay, but you can go into the options and check off "Disable Accidental Deletion Prevention". If you do this before trying any deletions, you will never get prompted, not even just once the first time.

That's a poor argument. In return, I could tell you "You can reject the deletion operation when accidentally trying to delete a root/any node". After all, you're warned. And if you're recklessly clicking in production, then it's a different type of a problem. It's not that an accidental delete keystroke or mouse click will delete things. You will be presented with a warning dialogue box:

image. 

You want safety, switch the focus to be by default on the <kbd>Cancel</kbd>. Done. No accidental deletions and unnecessary complexity you're introducing.

Just as you're pushing the line on having the feature to work your way, I'm doing the same from the perspective of those whose flow is going to be impacted. If hundreds (or thousands) that used it didn't have an issue and now they need to go and discover how to disable the feature, it's called "pissing off customers". 

Saying that there are other maintainers of this project. I'll be quiet and let others speak.

logiclrd:
Wow, that difference in resolution in that screenshot is weirdly disjoint, and I guess is just what happens when you run S.W.F code on a high DPI monitor :-P Maybe another PR is needed to convert Service Bus Explorer to WPF. :-D

paolosalvatori:
Sorry for coming later to the discussion, but as @ErikMogensen knows well, I’m extremely busy working at a PoC in this period with a large account here in Europe, this is why I look at PRs and issues once every 3-4 days. I’d prefer to avoid to open up any input dialog and allow someone to enter the name of an entity or use any other convention to delete an object. I think it’s sufficient what the tool provides today, that is, right-click an entity, select delete and click OK in the confirmation dialog. Now if you right-click the namespace node and you select delete entities from the context menu, entities don’t get immediately deleted. A dialog opens and asks for a confirmation. I think that this measure is enough. I think that introducing an input box to specify entities to delete using any sort of formalism, doesn't solve the problem, because people could make mistakes even when using this mechanism :) If you have a namespace with hundreds of queues, and you just want to delete a few of them, you can use the connect form to specify a filter to display only these entities and then you can right-click them and delete them. I think this is enough unless you have some special case that is not captured by these behaviors. Or am I missing something?

logiclrd:
> I think it’s sufficient what the tool provides today, that is, right-click an entity, select delete and click OK in the confirmation dialog. Now if you right-click the namespace node and you select delete entities from the context menu, entities don’t get immediately deleted. A dialog opens and asks for a confirmation. I think that this measure is enough. [..] Or am I missing something?

Just to be really explicit about the scenario I am trying to guard against, I frequently use Service Bus Explorer via a remote desktop tool. These tools send keystroke sequences like Ctrl-Alt-Delete, since of course this is just how Windows works, and on more than one occasion, I have unlocked a workstation running Service Bus Explorer via remote desktop to find the "delete everything" prompt already up, with the "OK" button focused. (I can only speculate that somehow the remote desktop tool's Ctrl-Alt-Delete ended up sending a Delete to Service Bus Explorer that it shouldn't have received.) At that point in time, an accidental space or enter keystroke would commence a deletion that I _really_ don't want happening.

It is extremely alarming to find the tool _already_ at this point, perched on the edge of disaster. If this is a scenario that doesn't come up for you, then I can see how it might be hard to identify with. :-)

I am hoping that for people who haven't been in this situation, being able to permanently shut off this feature with a single option checkbox is an acceptable trade-off for the peace of mind of people in my situation.

SeanFeldman:
Breaking my silence as the elephant in the room is continuously ignored. Is the problem that accident deletion can take place because of the incorrect focus on the <kbd>OK</kbd> button? If so, would the suggestion made in https://github.com/paolosalvatori/ServiceBusExplorer/pull/437#issuecomment-606871914 work?

In case the answer is "yes" - easy to address. If the answer is "no", elaborate on what's the problem.

I'm sorry, perhaps I'll come off as arrogant and impatient, but the amount of energy spent on this discussion could have probably fix the focus issue and provide the safety you're looking for.

logiclrd:
The problem _for me_ is that a dialog I didn't intentionally bring up is one errant action away from disaster. Changing the default button reduces the number of possible errant actions but is _still_ a dialog that is one click away from deleting things. Also, if the dialog receives a Tab keystroke, then it's right back to the OK button having focus.

In my mind, switching the default focus is a band-aid that doesn't fix the underlying problem and I would still be extremely alarmed to see the dialog come up unintentionally.

Am I right in understanding that you think Microsoft's own approach of requiring you to type the name of an ASB namespace in order to delete it is also overkill??

paolosalvatori:
I tend to agree with @SeanFeldman here, we can solve the problem of unattended deletion of entities just switching the default focus from the OK to the Cancel button. As an alternative, you can implement the following behavior:

1. - in case a user accidentally right-clicks the namespace and select delete entities, we can replace the current confirmation dialog box with an input box where you have to enter the name of the namespace.
2. - For all the other cases, I would keep the current confirmation dialog box. We can eventually switch the focus from the Ok button to the Cancel button.
3. - We can control the behavior at point 1 via a checkbox in the options form, but I would keep the current behavior as the default.

I’m going to copy and paste this reply in GitHub as well. If you agree with this plan, please go ahead and make the necessary changes. Please create an input box that complies with the look and feel of the DeleteForm or directly use the ParameterForm that you find in the ServiceBusExplorer project under Forms, thanks!

logiclrd:
This PR already does pretty much exactly those points.

- The prompt is integrated into `DeleteForm` but isn't imposed on every deletion. The call site showing the `DeleteForm` has to explicitly opt into it.
- All _mass_ deletions activate the function, and all regular deletions do not -- so regular deletions have a completely unchanged UI flow.
- The text that must be entered is the name of what you're deleting.
- The prompt has a checkbox beneath it that is also exposed in `OptionForm`.

The only difference is that if you haven't configured it at all, it will prompt you. I suspect that if the default is that it won't prompt you unless you explicitly ask for it, 99% of users won't even know the feature exists and won't know to go turn it on. Conversely, people who don't want the prompt will have to deal with it at most once, and have the option of using the options form to disable it without ever having entered the prompt text once.

You asked for a post including screenshots and describing the flow and if you scroll up you'll see one, it shows what the `DeleteForm` looks like with the new function enabled. The code has changed since the screenshot was taken, though; it now uses the name of the ASB namespace instead of the text `Absolutely Everything`.

jklaus:
I am also regularly seeing developers accidentally delete queues in our lower level environments.  I would greatly appreciate the default behavior be modified such that deleting isn't allowed unless specifically enabled.  Just having a confirmation on delete isn't sufficient as folks blow right past single click confirmation screens.  

Either that, or hide the default functionality deeper down in the context menu.  Maybe make a sub-context menu that pops out independently off the normal right click context menu.  Though my preference would be to modify the default behavior such that you have to enable deleting to begin with.  It'd be a one-time thing people have to do on a fresh install so it wouldn't be terribly annoying.  You could even still have the delete disabled in the context menu with a tooltip that states deleting must be enabled.. and probably state explicitly that delete is for deleting the queue.

SeanFeldman:
I can see an option to disable delete entities options/menus for the environments where modifications should be prohibited by having a special setting, but would not want by default jump through loops and hoops because of a few that don't exercise caution when in higher environments. It should be an opt-in option, not opt-out.

logiclrd:
@SeanFeldman Would you really be that affected by having to put a tick in a checkbox once, ever, after installing a new version? I don't understand your intense opposition to a fundamental safety feature.

paolosalvatori:
@SeanFeldman @logiclrd @jklaus I think we can proceed as follows: you can create an option as a checkbox or option buttons in the Options form. You can call this option safe massive delete with a confirmation. This option should be un-selected by default, and selected only on demand based on the guidance that developers receive from their managers. As an alternative, you can select this option, but only for massive delete operations that affect multiple entities. In this case, the user should enter the word 'yes' in an input box and click OK. This is the behavior followed by the Azure portal when a user tries to delete one or more resources from a resource group. Can you please implement any of these options and resubmit the PR? Please, make sure to include pictures to explain the new flow(s). Thanks!

jklaus:
*nudge*

logiclrd:
I _really_ want this feature to be enabled by default. In my mind, the trouble of turning the feature off just once forever is _way_ outweighed by the trouble of accidentally deleting important resources, regardless of the environment, and the way this is currently implemented, you can turn it off without even having to go through the prompt and type the response text -- it's just a checkbox in the options dialog. That only has to be done once. If you want to take what I've written and neuter it by having it completely invisible by default, I guess you can do that. :-(

It is curious to be that @paolosalvatori 's response suggests he hasn't seen that this pull request already adds:

- An option as a checkbox in the Options form
- Affects only massive delete options that affect multiple entities
- Requires the user to type in some short text in the input box, along the same lines as what Microsoft themselves do in Azure

The wording of your response @paolosalvatori  suggests that you don't think any of this has been implemented yet.

logiclrd:
Oops, how did I do that. Sigh, accidentally merged `master` in as well as `develop`. Fixing.

SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.2.0-PreventAccidentalMassDeletions.74.build.704 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/32832340) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/e9221317e2 by @logiclrd)

SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.2.0-PreventAccidentalMassDeletions.72.build.705 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/32832698) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/3b57ecfddf by @logiclrd)

ctoph13:
this feel's like an opt-out feature much more than an opt-in feature to me. it feels like familiar behavior when some freshly installed application has a confirmation dialog, and you go into settings to disable it if you don't want it to confirm every time anymore. 

+1 to having it enabled by default.

SeanFeldman:
@ctoph13, @jklaus, @logiclrd using your own argument, it's a one-time thing. If you care so much about it and allow this tool that requires **management** permissions to be used in production, would that be a deal-breaker if you get the feature disabled by the default, turn it on and continue from that point on?

Thousands of happy users are not chiming in on this issue, but you're willing to twist their experience to satisfy your requirement. Keep that in mind.

logiclrd:
Here's a possibility: Make it possible to disable with an environment variable, and then you can use Group Policy to push out that environment variable to all your machines and never see the prompt.

logiclrd:
Our argument is not that it's a one-time thing either way. Our argument is that as an opt-out feature, it's a one-time thing. As an opt-in feature, it's a zero-time thing. The majority of people won't know that it even exists.

jklaus:
The issue with opt-ing in to disable the delete is then your average el-cheapo developer isn't going to take the time to opt-in to disabling deletes.  I know that my argument should be solved by hiring proper engineers but there are many short sighted companies out there that think any warm body is capable of engineering software.  If this group of folks isn't willing to put this in place then I'll just fork it and do it in my version, whenever I can find the time.

logiclrd:
@jklaus FWIW, this PR has a complete implementation with no merge conflicts right now.

delroyh:
In my unsolicited opinion, this feature should be on by default. I have personal experience with team members inadvertently deleting resources via SBE that was 100% unintentional (I don't recall if this specific change would have solved all issues, but it speaks to the dangerous flexibility of SBE).

When we're building our products and infrastructure that utilizes ASB, we strive to build resilient and robust platforms. What this results in is platforms that don't often need human intervention. Why does this matter? Because it means that when we DO need to intervene, we're using SBE to perform operations infrequently and these operations aren't second nature given the lack of their occurrence.
It's been mentioned that this change will interfere with the work flows of developers frequently using the tool, but it is ALSO true that it will protect the work flow of developers and support engineers who infrequently use the tool to trouble shoot and resolve issues with production resources.

In fact, the ease with which it is currently possible to delete resources inadvertently is the cause of explicit warnings in my team onboarding documentation, and my explicit verbal warning every time I tell someone they "need to use Service Bus Explorer". In addition to the feature in discussion we should also be addressing the juxta-position of the resource "delete" button when viewing an individual message in a queue or topic, but I digress.

The statement from @SeanFeldman is fair: "Do not hurt others" but I can name dozens of industry standard products that have changed and as such impacted my workflow. The existing product has hurt us at it IS, and it is my opinion, that rebuilding a production queue and manually refilling it with messages you need to build by perusing through log messages is FAR more painful than having every developer check a box that says "Don't ask me again" one time.

My $0.02

Greybird:
Hi,

Just for the sake of users feedback, we regularly have issues with accidental deletions for users of this tool (I think some component focus switch issues which might have been already solved have made the problem more likely in the past)

It happens more often on non production environments, but can disrupt quite a lot of things also (developers/QA/... production environment are actually the non-production envs 😃).
I must also admit that a few months back, I did start the deletion of a production namespace queues, which I was able to stop by killing the app quickly, Damages were kept minor, but it could have been far worse.

IHMO, making the feature opt-in does not allow to get the full benefits of the feature.
It requires that people are aware that they will do mistakes before actually doing them. I think that being annoyed by the confirmation, and purposely changing the feature switch if the annoyance is too high is a way to ensure that people act on purpose, knowing the risks.
Of course this position is shaped by my experience, which has basically to cope with the fact that humans make errors, and, that as a human, I also do a lot of them, and having a safety net is something I personally prefer.

My $0.02 cents too, with ❤️ to both sides of the discussion!

ErikMogensen:
**Pro default confirmation**
@ctoph13 
@jklaus 
@logiclrd 
@delroyh 
@Greybird 
@ErikMogensen 

**Against default confirmation**
@SeanFeldman 
@paolosalvatori 

I haven't seen this engagement before on any issue so even though you are the author @paolosalvatori, I think you should change your mind on this.

SeanFeldman:
@ErikMogensen, no need to corner @paolosalvatori. I'll re-iterate again and repeat what I've said [earlier](https://github.com/paolosalvatori/ServiceBusExplorer/pull/437#issuecomment-624666745):

> I am not opposed to the feature. I do not share the same view as you about how it's implemented. Two different things.

The popup now allows users to opt-out of this confirmation flow. As long as it's persisted (which it seems to do), I'm fine.

paolosalvatori:
@ctoph13 @jklaus @logiclrd @delroyh @Greybird @ErikMogensen @SeanFeldman I'm not against the option, I'd like just to make a survey and understand what if SBE users prefer to configure this feature as an opt-in or opt-out feature. It comes without saying that most of the people that participated in the conversation in this PR are a small, negligible fraction of the total number of users. My proposal is to use a web site like GH Polls to ask feedback to the community about this setting this feature as an opt-in or opt-out feature. I want to remember everyone that I'm the author of this tool and I use a democratic approach and hear about everyone's opinion, but this needs to be posed in an educated, non-conflicting way.

logiclrd:
Soliciting feedback on a new feature for Service Bus Explorer. The new feature displays a prompt that requires the user to type a short amount of text to confirm that they really want to do bulk deletions. This is in order to prevent accidental bulk deletions. Whether this prompt is shown is configurable in the Options. This poll is about whether, on a fresh installation, the prompt should start out enabled or disabled. If the prompt starts out disabled, the only way to get a prompt is to know that the feature exists and explicitly, manually turn it on.

[](https://api.gh-polls.com/poll/01E8CEYKC8QSJ23RV55PY64K7V/Prompt%20should%20start%20out%20enabled%2C%20option%20to%20disable/vote)
[](https://api.gh-polls.com/poll/01E8CEYKC8QSJ23RV55PY64K7V/I%20should%20only%20get%20a%20prompt%20if%20I%20explicitly%20turn%20it%20on/vote)

paolosalvatori:
According to the result of the poll, I approve this PR. Merging is blocked by a @SeanFeldman request for change. I could squash and merge the PR, but I'll let Sean decide upon his request :)

Issue #464, opened 2020-09-01
Feature Request: Add Queue Listener for specific Message Session


jcoeltjen:
Is it possible to add a queue listener for a specific message session?
Or more specific: Is it currently possible to register a message handler for specific sessions or as a fallback register a session handler together with respective handlers?

We use session enabled queues for handling semi synchronous request/response calls for some legacy applications and have currently some problems in production and would like to peek some messages that belong to only one of the running message sessions.

My workaround for that right now is to use a custom C# project for that but it would be really nice to have this functionality inside the Service Bus Explorer.
If I am not alone with this feature request I would try to take the time to prepare a PR for this as a base for further discussion.

paolosalvatori:
Do you have the bandwidth to implement this feature and submit a PR? Thanks!

jcoeltjen:
> Do you have the bandwidth to implement this feature and submit a PR? Thanks!

I will give this a go!

jcoeltjen:
I tried a minimalistic approach on implementing this feature and opened a PR for this (#470).
As I am a little bit rusty in writing code for WinForms please **don't** be gentle in your review.

paolosalvatori:
Take your time @jcoeltjen no rush, and thanks for the contribution :)

ErikMogensen:
Fixed in #470

PR #470, opened 2020-10-14
Feature: Retrieve queue messages from a specific session only

jcoeltjen:
# Reason for the PR
This PR implements the feature request #464 in a very minimalistic and non-invasive way.

# Scope / Background
Adds support for limiting retrieving messages to specific sessions only.
Currently only Queues are supported but Session aware Subscriptions should also work with minimal extra work.
For this I will create another small PR if this one is accepted.

I tried to touch as less code as possible to implement this.

In this version the user has to manually type in the Session ID in the dialog window for retrieving messages.
If requested a list of available messages sessions could be provided in the dialog aswell but then I would prefer to extract the dialog and implement one Session-aware dialog and one for all non-session-aware entities.

image

### DIFF
--- a/src/ServiceBusExplorer/Controls/HandleQueueControl.cs
+++ b/src/ServiceBusExplorer/Controls/HandleQueueControl.cs
@@ -356,7 +356,7 @@ namespace ServiceBusExplorer.Controls
         {
             using (
                 var receiveModeForm = new ReceiveModeForm(RetrieveMessagesFromQueue, MainForm.SingletonMainForm.TopCount,
-                    serviceBusHelper.BrokeredMessageInspectors.Keys))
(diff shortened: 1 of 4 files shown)


SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 4.2.0-464-feature-message-sessions.75.build.720 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/35756615) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/70abb81e6c by @jcoeltjen)

ErikMogensen (approved):
Looks good to me. Thanks!

paolosalvatori (approved):

PR #489, opened 2020-12-30
Prevent tab to switch after receiving messages

Greybird:
Prior to this PR, when you used the Receive messages features and you used the `Enter` key to close the modal form allowing to provide details (mode, number of messages), quite frequently, immediately after receiving messages, the right tab switched back to the queue/subscription details.
Apart from the annoyance, when using Receive mode, it meant messages were lost without a way to get them anymore.

This was due to the fact that events are sent in the following order:
* `KeyDown`
* `KeyPress`
* `Keyup`

As the `KeyPress` was consumed to close the modal form, the `KeyUp` was sent to the parent window, which is often in this case the tree view, in turn resulting in its `KeyUp` event to fire.

Switching to `KeyDown` for the tree view easily solves this, and makes the behavior consistent with the error list view.


SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 5.1.0-bugfix-tab-change.2.build.739 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/37043648) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/3f7dc7ed72 by @Greybird)

ErikMogensen (approved):
Smart solution

SeanFeldman (approved):

paolosalvatori:
Thanks @Greybird @SeanFeldman @ErikMogensen

Issue #527, opened 2021-03-29
.NET support


SeanFeldman:
In the past, converting to .NET Core was a substantial effort. With .NET 5 the migration story has improved and should be somewhat easier with the[ .NET Upgrade Assistant](https://docs.microsoft.com/en-us/dotnet/desktop/winforms/migration/?view=netdesktop-5.0) (tool: https://dotnet.microsoft.com/platform/upgrade-assistant).

This could be the first step towards a potential cross-platform SBE.

SeanFeldman:
I've quickly tried to run it on `ServiceBusExplorer.csproj` but wasn't successful. Could be I've missed something. 
![image](https://user-images.githubusercontent.com/1309622/112866400-fc65c200-9076-11eb-8c73-772ea14ea3c6.png)

SeanFeldman:
I've tried to run it locally on `ServiceBusExplorer` and `ServiceBusExplorer.Tests` and the biggest blocker was `WindowsAzure.ServiceBus` (see [log.txt](https://github.com/paolosalvatori/ServiceBusExplorer/files/6223101/log.txt) for details).

PR #528, opened 2021-03-29
Upgrade to .NET 5

SeanFeldman:
Related to #527


SeanFeldman:
@paolosalvatori, @ErikMogensen, this was surprisingly simple and the outstanding issues are not that big based on my quick evaluation. The devil is in the details. If you'd like to have a swing at this PR to see if we could bring it to completion, great. Otherwise, it's just a pointer you can close.

SeanFeldman:
:x: [Build ServiceBusExplorer 5.1.0-net5-migrate.22.build.808 failed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/38457191) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/b396072440 by @SeanFeldman)

SeanFeldman:
WCF is gone from .NET Core/5. The legacy ASB (track 0) SDK is another one that would need to be addressed.

paolosalvatori:
@SeanFeldman @ErikMogensen we can safely remove WCF from the code and as an option from the message format dropdownlist. Those who still need to use WCF can use an older, legacy version. In other words, we can use the same approach that we followed when the Service Bus team dismissed Service Bus for Windows Server: we created a special release, interrupted the backward compatibility across the on-premises and cloud version of Service Bus, and added a link to the special release to the README file. If you manage to make it work @SeanFeldman we can migrate to .NET 5 which would provide multi-platform support.  Let me know if and how I can help here. We can have a call one of these days to check the current state of the migration and decide how to move this forward.

razvangoga:
Hi, if you need help on this, I'm willing to contribute some time in the next few weeks - I have a few macusers on my team that could benefit from a cross platform version

ErikMogensen:
Excellent @razvangoga! Unfortunately even if all of it is migrated to .NET 5 it will only run on Windows due to the dependency on  WinForms. I hope you are still willing to help with this.

My personal vision is that In a far future, the non-GUI code has have been moved out of ServiceBusExplorer.exe and into the other assemblies and have a x-plat GUI running on these assemblies. X-plat is the most asked for feature.

razvangoga:
Hmm.. bummer :( but sure, I'm still open to helping

serbrech:
even a cross plat cli on top of these library would be a nice start

PR #543, opened 2021-05-14
Log exception when listing entities in the namespace

bjorn-ali-goransson:
Hello,

I'm dealing with a SB connection where we do not have access to list subscriptions or topics.

This doesn't really work, but an initial hurdle is that establishing the connection (specifically listing the subscriptions and "event hubs") seems to get into some kind of deadlock. At least now I get an exception.

image


SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 5.1.0-add-async-to-listing-methods.27.build.834 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/39166019) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/0953e8649c by @bjorn-ali-goransson)

bjorn-ali-goransson:
Perhaps a solution to my ultimate problem would be to manually enter the subscription and topic here, in two new fields?

(I'm not saying it will result in a good UX 😆)

ErikMogensen (approved):
This is useful for other changes too.

ErikMogensen:
I recognize this problem. I think having another type of Connect would  work. Perhaps one menu item called "Connect to namespace" and another called "Connect to entity", where you only get access to one entity.

SeanFeldman (approved):

SeanFeldman:
Thank you @bjorn-ali-goransson.

paolosalvatori:
Thanks for the PR @bjorn-ali-goransson ❤️

bjorn-ali-goransson:
Thank you back for making this awesome tool!

PR #550, opened 2021-05-20
Update TimeSpanControl so that if any field is filled, others are defaulted to zero

### LINKED ISSUE #548: "field must be a number" when creating queue/topic entity
Hello, 

I noticed in the newer versions when creating a messaging entity like queue or topic, and not filling in all of the fields it cannot create due to validation "The X value of the X field must be a number". It seems it was a good approach to default to 0 when a value is not provided. But perhaps there is some thought behind adding this validation ? It's not an issue, but just a small inconvenience :)

Tomasz

Greybird:
This fixes #548, and should resume same behavior as in 4.1.112:
* not entering any value in any field will lead to a null value being returned by the control, and the default value for the entity will be used
* entering a value in any field will default the unfilled ones to 0

### DIFF
--- a/src/ServiceBusExplorer/Controls/NumericTextBox.cs
+++ b/src/ServiceBusExplorer/Controls/NumericTextBox.cs
@@ -82,19 +82,47 @@ namespace ServiceBusExplorer.Controls
 
         public bool IsFilled => !string.IsNullOrWhiteSpace(Text);
 
-        public bool IsValidIntegerValue => int.TryParse(Text, out _);
+        public bool IsValidIntegerValue => TryParseInt(out _);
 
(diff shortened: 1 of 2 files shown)


SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 5.1.0-allow-empty-timestamp.32.build.841 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/39242432) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/d23d565dc0 by @Greybird)

SeanFeldman (approved):

paolosalvatori:
Thanks @Greybird

PR #422, opened 2020-02-17
Fixed issue 375. Code was already working, just needed some bug fixes…

rajrao:
Fixes #375 
This looks like its already working correctly. I had to remove the entitySpecific check and also fix an issue with FilterOutException (which must have happened when the code was changed to be async). After that it works! The only thing is that it spits out an ugly error as it spits out the whole stack (not just the message). Thought of changing that, but the entire stack can be nice!

Tested both topic and queue based connection strings
Endpoint=sb://xxxx.servicebus.windows.net/;SharedAccessKeyName=test;SharedAccessKey=abcd;EntityPath=topicPath

PR #562, opened 2021-06-11
Remove check for non-empty Entity Path

### LINKED ISSUE #560: Cannot save entity specific connection string
My connection string includes an `EntityPath=...` stanza at the end.

Trying to save this string in the connection dialog displays an error message:
image

Removing the Entity Path setting and hitting "Save" stores the configuration.

But I cannot connect without the entity path being present, so every time I need to use `File -> Connect..`, enter the entity path and hit OK instead of just using the saved connection from `File -> Saved Connection` (ie. <kbd>Ctrl+1</kbd>).

avdv:
In the same vein as #422 this simply removes the check for entity specific connection strings since it is just working fine.

Fixes #560

### DIFF
--- a/src/ServiceBusExplorer/Forms/ConnectForm.cs
+++ b/src/ServiceBusExplorer/Forms/ConnectForm.cs
@@ -84,11 +84,6 @@ namespace ServiceBusExplorer.Forms
         //***************************
         private const string ConnectionStringCannotBeNull = "The connection string cannot be null.";
 
-        private const string ConnectionStringCannotBeEntitySpecific = "The connection string cannot be entity specific.";
-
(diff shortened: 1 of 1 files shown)


SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 5.1.0-save-entity-specific-path.37.build.849 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/39559975) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/706a75d84b by @avdv)

ErikMogensen (approved):
First PR I have seen that only removes code 😺

SeanFeldman (approved):

PR #591, opened 2021-09-26
Upated NuGet packages

### LINKED ISSUE #590: Receive and Delete behaviour can have side effects
During troubleshooting I found out how **Receive and Delete** works,  When I select a higher value for **Top** than current Queue length I received the messages and inspected. In the background new messages that arrive are Received and deleted until the top value is reached. The messages that arrive later will never get to the UI and will "disappear".  If you select ALL the loop will continue until the application is exited. 
I assume it is intentional but it can have side effects. In my case I did runs in batches and sometimes messages disappeared without explanation, until I got Enlighted. What happens if someone troubleshoots in production without knowing?

Now I know and aware of it, that said would it be better to change behavior to get up to current queue length?

ErikMogensen:
Updated the main NuGet packages to their latest versions. 

If we have extreme luck this fixes #590.

### DIFF
--- a/src/Common/Common.csproj
+++ b/src/Common/Common.csproj
@@ -15,11 +15,8 @@
     <DebugType>pdbonly</DebugType>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Azure.Core" Version="1.16.0" />
-    <PackageReference Include="Azure.Core.Amqp" Version="1.2.0" />
-    <PackageReference Include="Azure.Messaging.EventHubs" Version="5.5.0" />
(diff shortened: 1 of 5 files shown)


SeanFeldman (approved):

SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 5.1.0-update-nugets.54.build.880 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/40911226) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/663402c6da by @ErikMogensen)

PR #610, opened 2021-12-18
Set default queue or subscription when resubmitting messages

Greybird:
This small change aims to ensure that when resubmitting messages, the entity selector will have the entity from which messages have been initially received selected as default.
In my experience, it is the most common scenario when resubmitting messages to go to the same entity (from deadletter to active queue for instance), or to a related entity, which is often close to the one where messages have been received in the selector.

### DIFF
--- a/src/ServiceBusExplorer/Controls/HandleQueueControl.cs
+++ b/src/ServiceBusExplorer/Controls/HandleQueueControl.cs
@@ -3132,7 +3132,7 @@ namespace ServiceBusExplorer.Controls
             {
                 return;
             }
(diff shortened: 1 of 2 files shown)


SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 5.1.0-resubmit.72.build.899 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/41939063) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/c3f095d225 by @Greybird)

ErikMogensen (approved):
Simple, but quite an enhancement for the user experience!

PR #628, opened 2022-04-06
Fix null ref creating new subscription

ErikPilsits-RJW:
Fixes null reference exception when creating a new subscription, introduced by commit 09aa1eb5d1b1606935a08dc03f0f071e5508018c .


SeanFeldman (approved):

SeanFeldman:
It was not supposed to go to `master` 😞 
@paolosalvatori we need to ensure that contributor PRs can only target `develop`.
Something to review when @ErikMogensen moves to GH Actions.

PR #648, opened 2022-05-18
Support inconsistent schemes that have no impacts on actual behavior

Greybird:
Currently, relative urls for subscriptions forwards for example, are checked by ensuring the uri of the namespace is a base of the subscription url, before getting the relative path.
When the url is not absolute, or the condition above is false, the code fallbacks to a simple and imperfect string parsing code.

Recently, it has been put to my knowledge that some tooling where using different schemes for the urls (in this case, MassTransit uses `http://` rather than `sb://`).
This has no impact on the actual behavior of the service bus.

As such, I propose this PR to ensure the scheme is ignored when doing this matching, by copying the scheme of the namespace to the compared uri.
For technical purposes, I had to copy the port from the namespace also to prevent a default port to be added the the built url when there was none.

### DIFF
--- a/src/Common/Helpers/ServiceBusHelper.cs
+++ b/src/Common/Helpers/ServiceBusHelper.cs
@@ -5324,7 +5324,11 @@ namespace ServiceBusExplorer
         {
             if (Uri.IsWellFormedUriString(address, UriKind.Absolute))
             {
-                var uri = new Uri(address, UriKind.Absolute);
+                var uri = new UriBuilder(new Uri(address, UriKind.Absolute))
+                {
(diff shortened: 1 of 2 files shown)


SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 5.1.0-scheme.79.build.938 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/43583430) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/1dd1c1b817 by @)

SeanFeldman (approved):

ErikMogensen (approved):
Wonderful that you created unit tests too!

PR #660, opened 2022-06-29
Message filter - simple string contains

phatcher:
Not ready to be merged yet, but gives you an idea of approach...

1. Works on Queues/DLQ
1. Simple string.Contains on message content
    1. Option for regex?
    1. Should we apply before/after specified IBrokeredMessageInspector, option?
    1. Want to know how many accepted/rejected 
1. Refactored LogBrokeredMessageInspector to extract common code
1. Refactored IBrokerMessageInspector into a factory extension method to extract common code
1. Also apply to delete or presume we are already filtered?

### DIFF
--- a/src/Common/Helpers/BrokeredMessageExtensions.cs
+++ b/src/Common/Helpers/BrokeredMessageExtensions.cs
@@ -28,6 +28,8 @@ using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Text;
+using System.ServiceModel.Channels;
+using System.Xml;
 
 #endregion
 
@@ -35,7 +37,10 @@ namespace ServiceBusExplorer.Helpers
 {
     public static class BrokeredMessageExtensions
     {
+        public const string UnableToReadMessageBody = "Unable to read the message body.";
+
         private static readonly PropertyInfo BodyStreamPropertyInfo;
+        private const int MaxBufferSize = 262144; // 256 KB
 
         static BrokeredMessageExtensions()
         {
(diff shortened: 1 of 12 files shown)


SeanFeldman:
:white_check_mark: [Build ServiceBusExplorer 5.1.0-filter.82.build.952 completed](https://ci.appveyor.com/project/seanfeldman/servicebusexplorer/builds/44029616) (commit https://github.com/paolosalvatori/ServiceBusExplorer/commit/ee9e28f723 by @)

SeanFeldman:
Thank you, @phatcher.

> Not ready to be merged yet,
Making this a draft until it's ready.

src/Common/Helpers/BrokeredMessageExtensions.cs
```diff
@@ -28,14 +28,19 @@
...
 #endregion
 
 namespace ServiceBusExplorer.Helpers
 {
     public static class BrokeredMessageExtensions
     {
+        public const string UnableToReadMessageBody = "Unable to read the message body.";
+
         private static readonly PropertyInfo BodyStreamPropertyInfo;
+        private const int MaxBufferSize = 262144; // 256 KB
```
SeanFeldman, line 43:
```suggestion
        const int MaxBufferSize = 1024 * 256; // 256 KB
```

Alternatively, could be done in the following way:

const int MaxBufferSizeInKB = 1024 * 256;

src/Common/Helpers/BrokeredMessageExtensions.cs
```diff
@@ -51,6 +56,134 @@ static BrokeredMessageExtensions()
             }
         }
 
+        /// <summary>
+        /// Reads the content of the BrokeredMessage passed as argument.
+        /// </summary>
+        /// <param name="messageToRead">The BrokeredMessage to read.</param>
+        /// <returns>The content of the BrokeredMessage.</returns>
+        public static string GetMessageText(this BrokeredMessage messageToRead)
```
SeanFeldman, line 64: How about renaming this to `GetMessageBodyAsString` or `GetMessagePayloadAsString`? That's what this extension method is trying to do.

src/Common/Helpers/BrokeredMessageExtensions.cs
```diff
@@ -51,6 +56,134 @@ static BrokeredMessageExtensions()
...
+
+            if (messageToRead == null)
+            {
+                return null;
+            }
+            var inboundMessage = messageToRead.Clone();
+            try
+            {
+                stream = inboundMessage.GetBody<Stream>();
+                if (stream != null)
```
SeanFeldman, line 77: I'm a fan of early exit if possible and less identation. I.e.
```
if (stream is null)
{
  return null;
}
// the rest of the logic if stream is not null
```

src/ServiceBusExplorer.sln.DotSettings
```diff
@@ -589,6 +589,7 @@ II.2.12 &lt;HandlesEvent /&gt;&#xD;
 	<s:String x:Key="/Default/CodeStyle/Naming/XamlNaming/UserRules/=NAMESPACE_005FALIAS/@EntryIndexedValue">&lt;Policy Inspect="True" Prefix="" Suffix="" Style="aaBb" /&gt;</s:String>
 	<s:String x:Key="/Default/CodeStyle/Naming/XamlNaming/UserRules/=XAML_005FFIELD/@EntryIndexedValue">&lt;Policy Inspect="True" Prefix="" Suffix="" Style="AaBb" /&gt;</s:String>
 	<s:String x:Key="/Default/CodeStyle/Naming/XamlNaming/UserRules/=XAML_005FRESOURCE/@EntryIndexedValue">&lt;Policy Inspect="True" Prefix="" Suffix="" Style="AaBb" /&gt;</s:String>
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EdotCover_002EIde_002ECore_002EFilterManagement_002EModel_002ESolutionFilterSettingsManagerMigrateSettings/@EntryIndexedValue">True</s:Boolean>
```
SeanFeldman, line 592: Shouldn't be changed in this PR

src/Common/Helpers/BrokeredMessageExtensions.cs
```diff
@@ -51,6 +56,134 @@ static BrokeredMessageExtensions()
...
+        /// <param name="messageToRead">The BrokeredMessage to read.</param>
+        /// <returns>The content of the BrokeredMessage.</returns>
+        public static string GetMessageText(this BrokeredMessage messageToRead)
+        {
+            string messageText = null;
+            Stream stream = null;
+
+            if (messageToRead == null)
+            {
+                return null;
```
SeanFeldman, line 71: Would returning an empty string work better for the caller?
I.e. it's safer to operate on and might remove null checks.

SeanFeldman (commented):
A few minor things for your consideration.

src/Common/Helpers/BrokeredMessageExtensions.cs
phatcher: That's what the original method in LogBrokeredMessageInspector was called so I retained it - happy to change.

src/Common/Helpers/BrokeredMessageExtensions.cs
phatcher: Happy to change the calc - kept it as private as was in `LogBrokeredMessageInspector` and not used elsewhere so not sure about increasing the scoped

src/ServiceBusExplorer.sln.DotSettings
phatcher: ok

src/Common/Helpers/BrokeredMessageExtensions.cs
phatcher: Original code - need to check impact on `LogBrokeredMessageInspector`

SeanFeldman:
@phatcher are you still working on this PR?

PR #667, opened 2022-07-24
Use GitHub actions for CI/CD

### LINKED ISSUE #598: Switch to using GitHub Actions for releases
Switch to using GitHub Actions instead of AppVeyor and GitVersion. Use manually set tags for setting version.

ErikMogensen:
Fixes #598 

Replaces Appveyor and GitVersion with GitHub Actions. 

It is using an ugly way of verifying that releases are only done from the master branch, see https://github.community/t/how-to-create-filter-on-both-tag-and-branch/16936/2). 

I intend to wait with the publishing to Chocolatey until it has been verified that the publishing to GitHub works properly.

Since this is quite a change to the CI/CD system I'll wait for the go ahead from @paolosalvatori  and @SeanFeldman before I make it ready for review.

Edit: Added branch check to the code since the initial PR.


paolosalvatori:
I don't have the bandwidth to check that everything works as expected and you are more knowledgeable than me in this area, so I trust your work and approve the PR 😄 @ErikMogensen @SeanFeldman

paolosalvatori (approved):

SeanFeldman:
@ErikMogensen it's a draft because it's still a WIP PR or is it ready to be reviewed? Thanks.

ErikMogensen:
@SeanFeldman, it is ready for review. I have now put it in 'Ready for review'.

.github/workflows/build-test.yml
```diff
@@ -0,0 +1,77 @@
...
+  BUILD_CONFIGURATION: Release
+
+jobs:
+  build_test_publish:
+    name: Build and run unit tests
+    runs-on: windows-2022
+    steps:        
+    - name: Download and install NUnit console runner
+      run: |
+        $nunitVersion='3.15.2'
```
SeanFeldman, line 27: You could simplify this part by installing a global .NET tool for [NUnit](https://www.nuget.org/packages/NUnit.Console) and removing the need to specify the version and MSI installation.

.github/workflows/handle-tag.yml
```diff
@@ -0,0 +1,69 @@
...
+        ref: master
+
+    - name: Get latest master commit hash
+      id: master-commit-hash
+      run: |
+        hash=$(git log -n1 --format=format:"%H")
+        echo "::set-output name=master-hash::${hash}"
+        echo $hash
+
+    - name: Exit if tag commit hash don't match master commit hash
```
SeanFeldman, line 33: I would suggest using a shorter `name` and using the detailed text in `description`.

Example:

```yaml: 'tag-not-found'
- name: 'Exit if tag commit hash don't match master commit hash'
- description: 
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,79 @@
...
+      id: get-cached-build
+      uses: actions/cache@v3.0.5
+      with:
+        path: src/ServiceBusExplorer/bin/Release
+        key: cached-output-${{ github.sha }}
+
+    - name: Did we get a cache hit
+      if: steps.get-cached-build.outputs.cache-hit != 'true'
+      run: |
+        echo "The cached output was not found"
```
SeanFeldman, line 35:
```suggestion
        echo "Couldn't find build artifacts. Expected to find artifacts in cache but no items were found."
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,79 @@
...
+        echo "NupkgFilename=ServiceBusExplorer.$($env:RELEASE_VERSION).nupkg" >> $env:GITHUB_ENV
+   
+    - name: Compress output
+      run: |
+        7z a $env:ZipFilename $env:OUTPUT_PATH\*.*
+        
+    - name: Create the release
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}  
+      continue-on-error: true
```
SeanFeldman, line 69: Is this right?

appveyor.yml
```diff
@@ -1,108 +0,0 @@
-init:
-- git config --global core.autocrlf false
-
-skip_tags: true
-skip_commits:
```
SeanFeldman, line 5: We should keep this part in the new flow. No need to run builds if static content files are changing.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,79 @@
...
+      run:  |
+        (Set-Content -Value "VERIFICATION=MD5" -Path $env:GITHUB_WORKSPACE\src\VERIFICATION.txt)
+      
+    - name: Generate verification file
+      run: .\Generate-VerificationFile.ps1 -Version $env:RELEASE_VERSION -TemplateFilePath VERIFICATION-Template.txt -ExePath "$env:OUTPUT_PATH\ServiceBusExplorer.exe" -OutputFilePath "$env:OUTPUT_PATH\VERIFICATION.txt"
+
+    - name: Modify .nuspec file
+      run: (Get-Content $env:GITHUB_WORKSPACE\src\ServiceBusExplorer\ServiceBusExplorer.nuspec).Replace("`$version`$", "$env:RELEASE_VERSION") | Set-Content $env:GITHUB_WORKSPACE\src\ServiceBusExplorer\ServiceBusExplorer.nuspec
+
+    - name: Pack Chocolatey
```
SeanFeldman, line 53: This is packaging Choco artifact, but where's the deployment? We'll need to have that step here as well, along with the secret set in GH secrets.

SeanFeldman (changes requested):
A couple of items for your review. Thanks.

.github/workflows/build-test.yml
ErikMogensen: You mean using nuget install as mentioned in https://stackoverflow.com/questions/67770726/run-nunit-test-in-git-actions-for-net-framework? 

I think it is better to use a specified version of NUnit since I don't see a problem of being a couple of releases behind the latest NUnit version, but hitting a regression bug in NUnit may take a long time to troubleshoot.

.github/workflows/handle-tag.yml
ErikMogensen: Sure. I'll change it to "Verify tag commit matches master commit".

.github/workflows/publish.yml
ErikMogensen: The continue-on-error is there to handle the case if the release has already been created to handle the case if you want to overwrite an existing release. It is fine by me to remove it though.

appveyor.yml
ErikMogensen: Unfortunately the pull_request event does not have a paths-ignore option, https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request so I don't see how that can be achieved. 

Tagging will not trigger the handling-pull-request workflow. It will trigger the handle-tag workflow though, but it will fail. As mentioned, it seems like it has to be that way.

.github/workflows/build-test.yml
SeanFeldman: You can lock into a specific version if you want. The main idea was not to use an MSI installer.

.github/workflows/publish.yml
SeanFeldman: You mean if the publish needs to be executed more than once, right? As long as Choco is idempotent should be ok.

.github/workflows/publish.yml
SeanFeldman: @ErikMogensen are you a registered user with Chocolatey? If you are, what's the public **username** you're using?

SeanFeldman: Is [that](https://community.chocolatey.org/profiles/erikmogensen) you, @ErikMogensen?

.github/workflows/build-test.yml
ErikMogensen: Great!

.github/workflows/publish.yml
ErikMogensen: Yes :)

SeanFeldman: 👍 
Invited

.github/workflows/publish.yml
ErikMogensen: Yes, or if the release has been previously created as a draft. I'll leave it like this for now.

.github/workflows/publish.yml
ErikMogensen: The deployment to Chocolatey is added now.

SeanFeldman (approved):

SeanFeldman:
Approved.

README.md will need to be updated to point to GitHub rather than AppVeyor.

ErikMogensen:
@SeanFeldman, is there anything we should do with Appveyor before merging this PR? Otherwise let's merge it. I'll create a separate PR for the Readme.md.

SeanFeldman:
AppVeyor project was deleted and won't interfere with GH Actions. Good to proceed 👍

PR #675, opened 2022-09-03
Enable a blank greeting

### LINKED ISSUE #588: "Hi mate, how are you?" cannot be changed and persisted
First of all - thank you for this _amazing_ tool! It's so super-useful.

My comment is a very trivial one, but I thought I'd raise it anyway. It's just about the "Hi mate, how are you?" text that's there by default when trying to send a new message. Azure Service Bus Explorer looks a very well done, professional-quality tool. And this might just be me, but that message just looks silly. When using it in a screen-share - I always feel I have to explain to other people in the call that it's not my message - it's for some reason the default message. It feels like an odd thing to have in an other-wise, very well polished application. It just feels totally out of place.

ErikMogensen:
Enable using a blank greeting. 

Fixes #588 - "Hi mate, how are you?" cannot be changed and persisted

### DIFF
--- a/src/Common/Helpers/MessageAndPropertiesHelper.cs
+++ b/src/Common/Helpers/MessageAndPropertiesHelper.cs
@@ -52,7 +52,6 @@ namespace ServiceBusExplorer.Helpers
         private const string Key = "Key";
         private const string Type = "Type";
         private const string Value = "Value";
-        private const string DefaultMessageText = "Hi mate, how are you?";
         #endregion
 
         #region Private Static Fields
@@ -146,15 +145,8 @@ namespace ServiceBusExplorer.Helpers
             out string messageText, out string messageFile)
         {
             messageText = ReadMessage();
-            messageFile = string.Empty;
+            messageFile = configuration.GetStringValue(ConfigurationParameters.FileParameter);
 
-            if (string.IsNullOrWhiteSpace(messageText))
-            {
-                messageText = configuration.GetStringValue(ConfigurationParameters.MessageParameter,
-                   DefaultMessageText);
-            }
-
-            messageFile = configuration.GetStringValue(ConfigurationParameters.FileParameter, messageFile);
             if (!string.IsNullOrWhiteSpace(messageFile) && File.Exists(messageFile))
             {
                 using (var streamReader = new StreamReader(messageFile))
--- a/src/Common/Helpers/MessageAndPropertiesHelper.cs
+++ b/src/Common/Helpers/MessageAndPropertiesHelper.cs
@@ -1,4 +1,4 @@
-﻿#region Copyright
+#region Copyright
 //=======================================================================================
 // Microsoft Azure Customer Advisory Team 
 //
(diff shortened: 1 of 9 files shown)


src/Common/Helpers/MessageAndPropertiesHelper.cs
```diff
@@ -70,15 +69,17 @@ public static void WriteMessage(string message)
         {
             try
             {
-                if (string.IsNullOrWhiteSpace(message))
+                if (null == message)
```
SeanFeldman, line 72: Is this intentional? It will accept now empty strings or a string with space.

src/Common/Properties/AssemblyInfo.cs
```diff
@@ -30,4 +30,4 @@
 //      Revision
 //
 [assembly: AssemblyVersion("1.0.0.0")]
-[assembly: AssemblyFileVersion("1.0.0.0")]
+[assembly: AssemblyFileVersion("1.1.0.0")]
```
SeanFeldman, line 33: Is this necessary?

src/ServiceBusExplorer/Controls/TestQueueControl.cs
```diff
@@ -352,16 +351,12 @@ private void InitializeControls()
...
-                                                     !string.IsNullOrWhiteSpace(mainForm.MessageText) ?
-                                                     mainForm.MessageText :
-                                                     DefaultMessageText,
+                                                     mainForm.MessageText ?? string.Empty,
                                                      txtMessageText);
 
-                txtLabel.Text = !string.IsNullOrWhiteSpace(mainForm?.Label) ?
-                                mainForm.Label :
-                                DefaultMessageText;
+                txtLabel.Text = mainForm.Label ?? string.Empty;
```
SeanFeldman, line 359: It feels like `SetFormattedMessage()` should be guarding against the null rather than the callers doing it everywhere.

SeanFeldman (commented):
A few questions/remarks

src/Common/Helpers/MessageAndPropertiesHelper.cs
ErikMogensen: Yes, that was the whole purpose of this PR. I think that allowing null should work for all of the messaging services but it would be hard to test.

src/Common/Properties/AssemblyInfo.cs
ErikMogensen: The Fileversion is set during the build so this may get confusing. I'll revert back to original. We should probably remove all FileVersions or set them to 0.0.0.0

src/ServiceBusExplorer/Controls/TestQueueControl.cs
ErikMogensen: Yeah, that would save some lines. On the other hand, the messaging services may have different handling of null values so we would have to have the null handling in the calling function. Also, I think it good to have it visible in the calling functions that it will switch from null to an empty string.

SeanFeldman: > On the other hand, the messaging services may have different handling of null values
I'm not sure what you mean. There are no messaging services at play other than Service Bus. And the service Bus doesn't care about the payload. Perhaps I'm missing something here. It feels error-prone to check for null and coalesce anywhere a nullable value is used instead of handling it in a single location. Which `SetFomatedMessage()` is essentially is.

SeanFeldman (approved):
I'll leave the decision about the duplicated null check to Erik and approve not to block the PR.

src/ServiceBusExplorer/Controls/TestQueueControl.cs
ErikMogensen: Maybe it depends which services you include in Service bus. I made the change to queues, topics, relay event hubs. I am not sure if Notification hubs was updated as well.

ErikMogensen:
I'll make the file version change in another PR.

PR #678, opened 2022-09-07
Removed description field from GitHub action

ErikMogensen:


.github/workflows/handle-tag.yml
```diff
@@ -30,8 +30,7 @@ jobs:
         echo "::set-output name=master-hash::${hash}"
         echo $hash
 
-    - name: Verify tag commit matches master commit
-      description: Exit if tag commit hash don't match master commit hash
+    - name: Verify tag commit matches master commit - exit if they don't match
```
SeanFeldman, line 33: Looks like we turned a name into a description 😂

SeanFeldman (approved):

PR #679, opened 2022-09-07
Upper casing secret reference 

ErikMogensen:
....even though I don't think it will help


SeanFeldman (approved):
It's a convention. Good to follow it 👍

PR #693, opened 2022-10-17
Fix the ability to connect using an EntityPath connection string

### LINKED ISSUE #611: Error when using SAS-policy and connection string with Service Bus Queue
When I use **RootManageSharedAccessKey** for the Service Bus Namespace I can see everything, see below
image

When creating a SAS policy (Manage, Send, Listen) on the queue only and connecting using the connection string:

- `<08:55:02> Failed to retrieve Service Bus topics. Exception: Microsoft.ServiceBus.Messaging.MessagingException: Cannot get entity 'my-queue' because it is not of type TopicDescription. Check that you are using method(s) with the correct entity type. ---> System.Runtime.Serialization.SerializationException: Error in line 1 position 2599. Expecting element 'TopicDescription' from namespace 'http://schemas.microsoft.com/netservices/2010/10/servicebus/connect'.. Encountered 'None'  with name '', namespace ''. `

- `<Error><Code>401</Code><Detail>User is not authorized. TrackingId:...-...</Detail></Error>`

image

When creating a SAS policy (Manage, Send, Listen) on the topic and working with subscriptions it seems to work:

image

Anything that I might have missed when connecting to queues?

 Thanks for assisting!

ErikMogensen:
Fixes #611

The EntityPath was previously removed from the connection string, since the Track 0 SDK cannot handle it. However, Track 2 SDK requires it. This PR preserves the original string.

The ConnectionStringContainsEntityPath method could have been put inside the IsPremiumNamespace method but having it outside makes it clearer that this is sort of a hack. To get a good solution I'd prefer to have an enum for the SKU, where one of the enum members should be "unknown" since it looks like we won't be able to get the SKU when using a connection string containing an EntityPath. In that case some parts of the GUI should be hidden.

But let's start with this and make sure it works.

### DIFF
--- a/src/Common/Helpers/ServiceBusHelper.cs
+++ b/src/Common/Helpers/ServiceBusHelper.cs
@@ -422,6 +422,19 @@ namespace ServiceBusExplorer
             }
         }
 
+        public string ConnectionStringWithoutEntityPath
+        {
+            get
+            { 
+                var builder = new ServiceBusConnectionStringBuilder(connectionString)
+                {
+                    EntityPath = string.Empty
+                };
+
+                return builder.ToString();
+            }
+        }
+
         /// <summary>
         /// Gets or sets the connection string.
         /// </summary>
@@ -677,7 +690,7 @@ namespace ServiceBusExplorer
             MessagingFactory factory;
             if (!string.IsNullOrEmpty(ConnectionString))
             {
-                factory = MessagingFactory.CreateFromConnectionString(ConnectionString);
+                factory = MessagingFactory.CreateFromConnectionString(ConnectionStringWithoutEntityPath);
             }
             else
             {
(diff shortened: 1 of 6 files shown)


src/Common/Helpers/ServiceBusHelper.cs
```diff
@@ -422,6 +422,19 @@ public string ServicePath
             }
         }
 
+        public string ConnectionStringWithoutEntityPath
+        {
+            get
+            { 
+                var builder = new ServiceBusConnectionStringBuilder(connectionString)
+                {
+                    EntityPath = string.Empty
```
SeanFeldman, line 431: Would `EntityPath` be empty if not specified by default? I.e. does it need to be explicitly set to an empty string?

SeanFeldman (approved):

src/Common/Helpers/ServiceBusHelper.cs
ErikMogensen: Yes, `EntityPath` is empty if it has not been specified so it does not explicitly not have to be set to an empty string. But then it needs to be tested if there is an `EntityPath` in it or have a field or property for storing that information.

PR #698, opened 2022-11-02
Support preview releases

### LINKED ISSUE #697: Support creating beta releases
We should be able to create beta/pre-release releases. These should be published to GitHub but marked as pre-release, but not published to Chocolatey.

ErikMogensen:
Fixes #697 

Since Assembly version and file version don't allow letters this PR adds setting the Informational file version, which allows letters. Then we can have 5.3.4-beta1.

### DIFF
--- a/GetNumberVersion.ps1
+++ b/GetNumberVersion.ps1
@@ -1,0 +1,20 @@
+# Verify it is a proper version
+param(
+    [Parameter(Mandatory=$false)]
+	[string]
+    $Version
+)
+
+Set-StrictMode -Version 3
+
+if ($Version -match "^\d+\.\d+\.\d+$") {
+	return $Version
+}
+
+if (-not ($Version -match "^\d+\.\d+\.\d+-.{1,12}$")) {
+	throw "The version must start with three numbers separated by dots. This may be " + `
+		"followed by a '-' and a word. That word may not be longer than 12 characters. For instance, 4.3.22-beta."
+}
+
+$parts = $Version -split "-"
+return $parts[0]
--- a/GetNumberVersion.ps1
+++ b/GetNumberVersion.ps1
@@ -2,14 +2,10 @@
 param(
     [Parameter(Mandatory=$false)]
 	[string]
-    $Version = '0.0.0'
+    $Version 
 )
 
 Set-StrictMode -Version 3
-
-if ($Version -match "^\d+\.\d+\.\d+$") {
-	return $Version
-}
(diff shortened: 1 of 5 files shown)


GetNumberVersion.ps1
```diff
@@ -0,0 +1,20 @@
...
+
+Set-StrictMode -Version 3
+
+if ($Version -match "^\d+\.\d+\.\d+$") {
+	return $Version
+}
+
+if (-not ($Version -match "^\d+\.\d+\.\d+-.{1,12}$")) {
+	throw "The version must start with three numbers separated by dots. This may be " + `
+		"followed by a '-' and a word. That word may not be longer than 12 characters. For instance, 4.3.22-beta."
```
SeanFeldman, line 16: > That word may not be longer than 12 characters

Not according to [SemVer](https://semver.org/).

ErikMogensen: In the FAQ part it says: 

> Does SemVer have a size limit on the version string?
> 
> No, but use good judgment. A 255 character version string is probably overkill, for example. Also, specific systems may impose their own limits on the size of the string.

Anyway, I changed it to use the impressive regex mentioned on https://semver.org.

SeanFeldman: That's what my comment was about. The 12 characters constrain seemed unnecessary 🙂

SeanFeldman:
It's not my intent to stall this PR. Just doesn't feel right to introduce this complexity. Have you excited something like MinVer[MinVer](https://github.com/adamralph/minver)? This problem has been solved with tools such as MinVer.

ErikMogensen:
I haven't used Minver. It seems to be a good tool, but it looks to me that it will add complexity in this case since we still have to check if it is a prelease to determine if it should be published to Chocolatey and it requires changing all of the project files and adding one more component. 

To make changes to the build is time-consuming since a build takes 5 minutes and requires some gitting. That's a great advantage to PowerShell scripts which is just F5 and then it is running and allows debugging.

SeanFeldman:
Manually tweaking the version and figuring out what it should be is complex. This is what MinVer and such already solved a long-time ago if you prefer to do it manually, by all means.

SeanFeldman:
> I am not going to spend more time on this PR, unless someone start paying me for it.

☝️🙂

ErikMogensen:
It is worth a try :wink:

SeanFeldman (approved):

Issue #700, opened 2022-11-11
Test topic using more than one file will still only take one file


wmmihaa:
## To reproduce
1. Right-click a topic and select `Send messages`
2. In the *Send messages to <topic>" dialog, select the *Files* tab
3. Click the `Select Files` button and select more than one file
4. Click the `Start` button

Only one file gets published

ErikMogensen:
To send both of the files you have to change the 'Message count' property on the 'Sender' tab to a value higher than 1.

PR #701, opened 2022-11-11
Fixed issue with test topic using more than one file

wmmihaa:
Related to Test topic using more than one file will still only take one file #700

### DIFF
--- a/src/ServiceBusExplorer/Controls/TestTopicControl.cs
+++ b/src/ServiceBusExplorer/Controls/TestTopicControl.cs
@@ -895,6 +895,7 @@ namespace ServiceBusExplorer.Controls
                                 HandleException(ex);
                             }
                         }
+                        messageCount = fileList.Count();
                     }
                     else // Brokered Message Generator Tab
                     {


ErikMogensen (changes requested):
Hi Micke! Long time!

Great that you found and reported this bug. However, it seems that this fix makes the number of sent messages equal to the number of selected files. 

Please also provide a fix for queues to keep the behavior in sync.

wmmihaa:
> However, it seems that this fix makes the number of sent messages equal to the number of selected files.

Would this not be the expected behavior?

ErikMogensen:
The expected behavior is to send the number of messages that are set in the 'Message count' property on the 'Sender' tab. Since this may be confusing there is text explaining this behavior on the tab where you select the files.

I also replied to the issue, since it does not seem to be a bug.

wmmihaa:
In my humble opinion the `Message Count` setting makes sense when using *Messages* but is confusing when it comes to *Files*. I've made some changes automatically updating the `Message Count` property to the number of selected files, while setting it back to `DefaultMessageCount` (1) if any of the other tabs are selected. 

I've made the update for both Queues and Topics

https://user-images.githubusercontent.com/3214367/201526958-89348184-efd7-4eb5-9e26-eda546737b2e.mp4

ErikMogensen:
> In my humble opinion
I have never seen you being humble 😄 

Nice video and thanks for the addition to the queue control. I see two problems though. One is that if you have set the `Message count` to a higher value than the number of files it will be lowered. That could be fixed by making sure that the `Message count`is never lowered. The other problem is that changing the `Message count` may not be expected. That could perhaps be solved by changing the text that is above the list box and mentioning this behavior. 

@paolosalvatori, @SeanFeldman do you have any input?

BTW, @wmmihaa were you able to open the controls in the Visual Studio designer? I am unable to do that. Most likely due to them inheriting from an abstract base class.

wmmihaa:
> One is that if you have set the Message count to a higher value than the number of files it will be lowered.
Yes I was considering pointing this out, but thought less of it. But yes, if the user has selected the *Message* tab and sets the *Message Count* to any value but one, he or she might find it confusing that it's reset to "1" after selecting the *Files* tab and then back to *Messages*.

However, I think the confusion has to be put in perspective to the relation of selecting five files and hitting *Send* and only one message is transmitted. Or what happens if I select five files and set the *Message Count* to 7?

Issue #702, opened 2022-11-15
BulkPurge NullReferenceException When Topic Has No Subscriptions


avalanchis:
### Summary:

Selecting any of the options under Topics -> Purge will cause a Null Reference Exception if a Topic exists with no Subscriptions.

### Expected behavior:

1. Select **Topics** -> **Purge** -> **Messages**, **Dead-letter Queue Messages**, or **All Messages** 
2. The appropriate purge function is performed on all topics.

### Actual behavior:

1. Select **Topics** -> **Purge** -> **Messages**, **Dead-letter Queue Messages**, or **All Messages**
2. The following message appears in the Log window:

     `<09:52:57> Exception: Object reference not set to an instance of an object.`

SeanFeldman:
Closed via https://github.com/paolosalvatori/ServiceBusExplorer/commit/bc288d24ca69f1a78a3739fae837967d11380ff7

PR #703, opened 2022-11-15
Modify BulkPurge to handle Topic with no Subscriptions

avalanchis:
Fix for Issue [702](https://github.com/paolosalvatori/ServiceBusExplorer/issues/702).

Modify subscriptionsExtractor  anonymous function so that it handles the case where tn.FirstNode is null by returning an empty Enumerable.


SeanFeldman (approved):

SeanFeldman:
Thank you, @avalanchis.

Issue #505, opened 2021-01-07
Enhance the .editorconfig file


ErikMogensen:
Add more items to the .editorconfig file. More about it at https://docs.microsoft.com/en-us/visualstudio/ide/create-portable-custom-editor-options?view=vs-2019.

This may conflict with the Resharper settings, but hopefully the editorconfig settings override.

PR #721, opened 2023-05-15
Extended 'Accidental Deletion Prevention' Verification

BruAlcaraz:
I have created an extended deletion Prevention to avoid accidental deletion of Queues, Topics and subscriptions.
This new option is is enabled by default and can be disabled in the Options menu

### DIFF
--- a/src/Common/Helpers/ConfigurationHelper.cs
+++ b/src/Common/Helpers/ConfigurationHelper.cs
@@ -220,8 +220,6 @@ namespace ServiceBusExplorer.Helpers
 
             resultProperties.DisableAccidentalDeletionPrevention = configuration.GetBoolValue
                 (ConfigurationParameters.DisableAccidentalDeletionPrevention, currentSettings.DisableAccidentalDeletionPrevention, writeToLog);
-            resultProperties.DisableExtendedAccidentalDeletionPrevention = configuration.GetBoolValue
-                (ConfigurationParameters.DisableExtendedAccidentalDeletionPrevention, currentSettings.DisableExtendedAccidentalDeletionPrevention, writeToLog);
 
             resultProperties.ProxyOverrideDefault = configuration.GetBoolValue(ConfigurationParameters.ProxyOverrideDefault, currentSettings.ProxyOverrideDefault, writeToLog);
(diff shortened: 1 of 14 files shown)


ErikMogensen:
How is this related to the existing setting? 
image

BruAlcaraz:
If the Disable Accidental prevention is true then the new one is also going to be disabled automatically. But if the  Disable Accidental prevention is false then you can disable just the extended part depending on user preference.

BruAlcaraz:
For the user, I made it visually clear by making the extended option true and disabling it when the original disable accidental option is also true.

ErikMogensen (changes requested):
Thank you! 

I think it will be confusing for some users what the different delete prevention alternatives will do. What do you think of combining them into one? 

Also, can you provide more information about which scenario(s) this PR solves? Preferably in an issue that this PR can refer too.  

Thanks!

BruAlcaraz:
Hi @ErikMogensen,

I have merged the options as you suggested. 

I don't know if there are any Issues reported as this was an issue that I encountered myself with some of my support engineers as they accidentally deleted the entire queue when trying to delete some dead-letter messages. 

These changes extended the delete verification to anywhere where a main Item is been deleted (Queue, Topic, Subscription) to avoid accidental loss of possibly hundreds of queued messages.
The following 3 actions are the affected ones

image
image

BruAlcaraz:
Hi @ErikMogensen what do you think now with the changes?

BruAlcaraz (commented):
Changes done

BruAlcaraz:
@ErikMogensen, @paolosalvatori, @SeanFeldman 
I haven't heard back from you guys in a while. Could you guys take a look at this PR?
I have had 2 of my support engineers deleting the Queue when trying to delete a selection of dead letters this is why I decided to suggest this change. 

Thank you in advance.

ErikMogensen (changes requested):
Great! 

Sorry about the delay. 

Just one thing. Please add curly brackets around the bodies belonging to the if statements, since that is the way it is usually done in this repo. Also, it may prevent someone from adding a deletion statement after such a body, not realizing it will be not be conditional.

BruAlcaraz:
@ErikMogensen Brackets added as requested.
Please let me know if it looks good now :)

ErikMogensen (approved):
Thank you @BruAlcaraz. 

We have revived #505 to set expected code style when making PRs.

SeanFeldman (approved):

Issue #712, opened 2023-03-18
Feature Request: Re-submit a single message with binary content created with Azure.Messaging.ServiceBus nuget package


KinNeko-De:
In the Service Bus Explorer you can already "repair and resubmit" messages from a dead letter queue.

The content of the service bus message is a byte array (in C#).
You can use Protobuf to serialize an object and put it into the message.

As soon as you Resubmit the message with the Service Bus Explorer, the content of the message is not parsable by protobuf anymore. The reason for that is that the protobuf message is converted to a string and from that back to binary content. This change the content of the message.

Requirement:
Service Bus Messages can be resubmitted if they contain binary (example protobuf) content.
The content of the message is not changed.

Repro:
I have created a repro for the azure service bus explorer because I requested the feature there also : https://github.com/KinNeko-De/repro-servicebusexplorer-protobuf

PoC:
I have created a fork and hacked a solution into the existing code to resubmit messages without changing the content. https://github.com/KinNeko-De/ServiceBusExplorer
I added a small checkbox in the existing dialog.
I do not think this should be the final solution, but i am not so familiar with WPF.

An alternative solution would be add a to a specific menu item like repairAndResubmitMessageToolStripMenuItem.

I am open to other suggestions but will need some more time to implement this :)

Priority:
The priority is not so high. Resubmit messages in production is not a common task. If so multiple messages have to be resubmitted.
The most common usage is manual testing during development. For that, I have automated tests that send the same message again.

Ps.:
I have currently no access to a service bus instance which makes it hard to test this.

SeanFeldman:
> I have currently no access to a service bus instance which makes it hard to test this.

A [free account ](https://azure.microsoft.com/en-ca/free/)  might help.

> I added a small checkbox in the existing dialog.
I do not think this should be the final solution, but i am not so familiar with WPF.
>
> An alternative solution would be add a to a specific menu item like repairAndResubmitMessageToolStripMenuItem.

The tool is handling NServiceBus old encoding of the messages w/o any UI changes. Maybe the way it was done is the way to handle this case. Not sure if that's possible. A byte array should stay a byte array w/o conversion. Is that conversation to string and back happening during cloning of the original message?

A PR is a good way to have that discussion and see the changes.

KinNeko-De:
> A [free account ](https://azure.microsoft.com/en-ca/free/) might help.
From Azure
```
Start with USD200* Azure credit
You'll have 30 days to use it
```
[...]

PR #722, opened 2023-05-15
Re-submit a single message with binary content created with Azure.Messaging.ServiceBus nuget package

KinNeko-De:
I found out that the issue discussed in #712 only occurs when Azure.Messaging.ServiceBus is used.
I added UnitTest to show that the issue does not occur if you use Windows.AzureService.
I failed to reproduce exactly the error. I added a test that somehow emulates the error behavior. I do have not enough knowledge of what was changed between the two APIs. Also, the current Protobuf Nuget Package does not support 4.6.2 so i had to add the Protobuf message as a byte array :(

I ended up adding an additional menu item to resubmit a single message without changing the content.
I reused the code from the resubmit of multiple messages.
But that is just a workaround.

I manually tested re-submitting in Azure ServiceBus Queue and Topic and in both cases DLQ-Messages.
I was not able to manually test Azure Event Grid because I have no test environment.

Finally, I updated NUnit3TestAdapter to the newest version to locally run tests.

### DIFF
--- a/src/ServiceBusExplorer.Tests/Helpers/ServiceBusHelperTest.cs
+++ b/src/ServiceBusExplorer.Tests/Helpers/ServiceBusHelperTest.cs
@@ -3,6 +3,8 @@
 using ServiceBusExplorer;
 using System;
 using NUnit.Framework;
+using Microsoft.ServiceBus.Messaging;
+using System.IO;
 
 #endregion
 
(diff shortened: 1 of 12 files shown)


ErikMogensen:
Please provide screen shots of the new GUI parts.

KinNeko-De:
Before:
servicebusexplorer_before

After:
servicebusexplorer

Multiple Messages are unchanged:
servicebusexplorer_multiple

ErikMogensen (approved):
Really appreciate that you created unit tests. Thank you!

Personally I think that the new Resubmit menu item should be on top of "Repair and resubmit".

KinNeko-De:
Hey..
I am not completely sure what I have to do now.
Do you wait for me to move the menu item to the first place?

ErikMogensen:
No need to do anything @KinNeko-De. I'd prefer if @paolosalvatori or @SeanFeldman also reviewed the PR.

KinNeko-De:
@ErikMogensen Thx :) I was confused because the bot marked my issue as stale and 'wontfix' even if there is an open pull request.
I have no problem to wait :)

PR #728, opened 2023-07-22
Removed the class TestControlBase and added the class TestControlHelper

ErikMogensen:
Some time ago I created a base class for the TestControl* controls. Unfortunately this prevented them from opening in design mode.

This PR removes that base class and puts that code in a new class called TestControlHelper which is contained in each of the TestControl* classes. This is not very elegant, but better than duplicating the code in all of the TestControl* classes IMHO. Hopefully more code will be moved into TestControlHelper in the future.


src/ServiceBusExplorer/UIHelpers/TestControlHelper.cs
```diff
@@ -0,0 +1,54 @@
...
+
+namespace ServiceBusExplorer.UIHelpers
+{
+    internal class TestControlHelper
+    {
+        #region Internal Constants
+        //***************************
+        // Formats
+        //***************************
+        internal const string ExceptionFormat = "Exception: {0}";
```
SeanFeldman, line 15: If the class is internal, all `internal` properties could be `public` as the intent is for those to be available. Unless that's not the intent, those should be marked as `private`. The same applies to the rest of the class methods and member fields.

ErikMogensen: Changed. 

https://ericlippert.com/2014/09/15/internal-or-public/ is a great read on the subject.

SeanFeldman (approved):

ErikMogensen:
Changed. https://ericlippert.com/2014/09/15/internal-or-public/ is a great read on the subject.

Issue #731, opened 2023-07-30
Wrong test made when saving two files configuration


robionekenobi:
Hi,

In TwoFilesConfiguration.cs, there is a bug when testing if files have been modified

Actually the test is  (at line 470)
`if (!updatedApp && !updatedApp)`
but must be
`if (!updatedUser && !updatedApp)`

Regards
Robert

ErikMogensen:
Fixed by #732

PR #732, opened 2023-07-30
Fix bug Wrong test made when saving two files configuration

robionekenobi:
Hi,

here my fix to correct the bug #731

Regards
Robert


ErikMogensen (approved):
Good catch!

PR #735, opened 2023-08-09
docs: fix formatting

MichaelCurrin:
Fixed broken italics formatting.

I'd also happily convert that to a table of two columns instead of bullet points.


ErikMogensen (approved):

ErikMogensen:
Thanks. 

Go ahead and make a table if you think that will be better.

Issue #738, opened 2023-08-24
Event Hubs - Events Custom Properties on Listener screen don't render - started occurring since after Release 5.0.4


minascasiou:
When 
- connecting to Azure Event Hubs 

Then
on the **Listener** screen, Events tab
- Events Custom Properties on don't render 

This started occurring since after Release 5.0.4

SeanFeldman:
@minascasiou, a [potential workaround](https://github.com/paolosalvatori/ServiceBusExplorer/pull/740) w/o downgrading the SDK is coming.

ErikMogensen:
@minascasiou, please try the brand new 5.0.18 release.

minascasiou:
Awesome, thank you Erik!
Works like a charm!

PR #740, opened 2023-09-06
Abstracting EventHubs EventData

### LINKED ISSUE #720: When testing in-built Send Event and Create Partition/Consumer Group Listener, message is sent and received yet the message does not appear
Hi Guys,
Appreciate the work you guys put into the explorer. 
It really does make life easy.
However, I found an issue while testing the in-bult send and listen function to view the message.

**[Summary]**
I tried sending, receiving and reading messages through the built-in function as I wanted to see if the messages are going through fine. The message is sent and received fine but, the issue with reading the messages that was sent.
SendEvent
image

The functions used within the explorer are `Send Event` / `Create Partition Listener`
image
image

**[Issue]**
The issue appears from `5.0.7`, the version I had, and also updated to newest `5.0.15`, and the issue is still appearing as well.

When sending the message I can see an Exception returning from the activity log.

`Exception : This object is read-only and cannot be modified. `
This message also appears on the `properties` of `Event System Properties` when viewed from the Listener popup console. 

error

Also you can `Event Custom Properties` is blank, where it should have the `machine name` and the `user name`.

**[TroubleShooting / Validating]**
I have validated that with version `5.0.3`, we are able to send and receive the test message and be able to read the test message sent. 
image

From this image, you can see that the `properties` is blank without the exception.
As well as, the `Event Custom Properties` has `machine name` and `user name` filled up. 

Could you please fix this issue?
Thank you in advance.

SeanFeldman:
Fixes #720 #738 #638 

A breaking change was introduced in the track 0 SDK `WindowsAzure.ServiceBus` somewhere after version `6.0.3`. As a result of that, it broke SBE ability to clone EventHub messages for UI purposes. This PR is a workaround for the issue to allow rendering EH messages, its custom and system properties, w/o throwing exceptions.

https://github.com/paolosalvatori/ServiceBusExplorer/assets/1309622/95ebead3-8968-403c-b5b6-30ce443253c5

Disclaimer: I have not tested all the permutations of what can be done with EH. I would appreciate it if someone could take a look to see if there are any undesirable side effects. This is merely an attempt to unblock those who need to be able to list and evaluate messages sent to EH.

### DIFF
--- a/src/Common/Abstractions/EventDataMessage.cs
+++ b/src/Common/Abstractions/EventDataMessage.cs
@@ -1,0 +1,50 @@
+﻿namespace ServiceBusExplorer.Abstractions
+{
+    using System;
+    using System.Collections.Generic;
+    using System.IO;
+    using Microsoft.Azure.Amqp;
+    using Microsoft.ServiceBus.Messaging;
+
+    public class EventDataMessage : IDisposable
+    {
+        readonly EventData eventData;
+        Stream stream;
+
+        public EventDataMessage(EventData eventData)
+        {
+            this.eventData = eventData;
+            stream = eventData.GetBodyStream();
+            Properties = eventData.Properties;
+            PartitionKey = eventData.PartitionKey;
+            SequenceNumber = eventData.SequenceNumber;
+            Offset = eventData.Offset;
+            SerializedSizeInBytes = eventData.SerializedSizeInBytes;
+            EnqueuedTimeUtc = eventData.EnqueuedTimeUtc;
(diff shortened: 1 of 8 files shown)


SeanFeldman:
@paolosalvatori or @ErikMogensen, could you review/approve?

@DelS-AGL was kind enough to give a [quick test run](https://github.com/paolosalvatori/ServiceBusExplorer/issues/638#issuecomment-1709286770) and caught one regression that is now fixed.

paolosalvatori (approved):

PR #746, opened 2023-10-22
Add the ability to cancel scheduled messages

### LINKED ISSUE #360: Adding "Cancel Scheduled message" feature
https://docs.microsoft.com/en-ca/dotnet/api/microsoft.azure.servicebus.queueclient.cancelscheduledmessageasync?view=azure-dotnet#Microsoft_Azure_ServiceBus_QueueClient_CancelScheduledMessageAsync_System_Int64_

ErikMogensen:
Fixes #360

This PR adds the ability to cancel scheduled messages on queues. The Service bus API does not support browsing scheduled messages on subscriptions. 

A new context menu item "Cancel Scheduled Message" has been added as shown below:
image

If multiple messages are selected it says "Cancel Scheduled Messages" instead:
image

If the _Disable Accidental Deletion Prevention_ is disabled, a message box will ask for confirmation. 

The rows in the grid will not update, nor will the message count for the queue in the tree view. That is something that may be added later.

### DIFF
--- a/src/ServiceBus/Helpers/CancelScheduledMessagesHelper.cs
+++ b/src/ServiceBus/Helpers/CancelScheduledMessagesHelper.cs
@@ -1,6 +1,11 @@
 ﻿using System.Collections.Generic;
 using System.Diagnostics;
+using System.Threading;
 using System.Threading.Tasks;
+
+using Azure.Messaging.ServiceBus;
+
+using ServiceBusExplorer.Utilities.Helpers;
 
 namespace ServiceBusExplorer.ServiceBus.Helpers
 {
@@ -10,23 +15,39 @@ namespace ServiceBusExplorer.ServiceBus.Helpers
             string queueName, List<long> sequenceNumbersToCancel)
         {
             var client = serviceBusHelper.CreateServiceBusClient();
-            var sender = client.CreateSender(queueName);
 
-            serviceBusHelper.WriteToLog($"Starting cancellation of scheduled messages on queue {queueName}.");
-
-            var stopwatch = Stopwatch.StartNew();
-
-            // Got lots of Azure.Messaging.ServiceBus.ServiceBusExceptions and the UI stopped working when testing 
-            // with 1000 messages when calling using Task.WhenAll. So, doing it one by one instead.
-            foreach (var sequenceNumber in sequenceNumbersToCancel)
+            try
             {
-                await sender.CancelScheduledMessageAsync(sequenceNumber);
-                serviceBusHelper.WriteToLog($"Cancelled scheduled message with sequence number {sequenceNumber}.");
+                var sender = client.CreateSender(queueName);
+
+                serviceBusHelper.WriteToLog($"Starting cancellation of scheduled messages on queue {queueName}.");
+
+                var stopwatch = Stopwatch.StartNew();
+                var semaphore = new SemaphoreSlim(40); // As recommended by https://learn.microsoft.com/en-us/azure/service-bus-messaging/message-transfers-locks-settlement#settling-send-operations
+                var tasks = new List<Task>(sequenceNumbersToCancel.Count);
+
+                foreach (long sequenceNumber in sequenceNumbersToCancel)
+                {
+                    await semaphore.WaitAsync();
+                    tasks.Add(CancelScheduledMessageWithLogAsync(sender, sequenceNumber, serviceBusHelper.WriteToLog)
+                        .ContinueWith((t) => semaphore.Release()));
+                }
+
+                await Task.WhenAll(tasks);
+                stopwatch.Stop();
+                serviceBusHelper.WriteToLog($"Cancelled {sequenceNumbersToCancel.Count} scheduled message(s) in {stopwatch.ElapsedMilliseconds / 1000} seconds.");
             }
+            finally
+            {
+               await client.DisposeAsync();
+            }
+        }
 
-            stopwatch.Stop();
-
-            serviceBusHelper.WriteToLog($"Cancelled {sequenceNumbersToCancel.Count} scheduled message(s) in {stopwatch.ElapsedMilliseconds / 1000} seconds.");
+        static async Task CancelScheduledMessageWithLogAsync(ServiceBusSender sender, 
+            long sequenceNumber, WriteToLogDelegate writeToLog)
+        {
+            await sender.CancelScheduledMessageAsync(sequenceNumber);
+            writeToLog($"Cancelled scheduled message with sequence number {sequenceNumber}.");
         }
     }
 }
--- a/src/ServiceBus/Helpers/CancelScheduledMessagesHelper.cs
+++ b/src/ServiceBus/Helpers/CancelScheduledMessagesHelper.cs
@@ -35,7 +35,8 @@ namespace ServiceBusExplorer.ServiceBus.Helpers
 
                 await Task.WhenAll(tasks);
                 stopwatch.Stop();
-                serviceBusHelper.WriteToLog($"Cancelled {sequenceNumbersToCancel.Count} scheduled message(s) in {stopwatch.ElapsedMilliseconds / 1000} seconds.");
+                Func<string> singleOrPlural = () => sequenceNumbersToCancel.Count() > 1 ? "messages": "message";
+                serviceBusHelper.WriteToLog($"Cancelled {sequenceNumbersToCancel.Count} scheduled {singleOrPlural()} in {stopwatch.Elapsed}.");
             }
             finally
             {
--- a/src/ServiceBus/Helpers/CancelScheduledMessagesHelper.cs
+++ b/src/ServiceBus/Helpers/CancelScheduledMessagesHelper.cs
@@ -46,8 +46,10 @@ namespace ServiceBusExplorer.ServiceBus.Helpers
         static async Task CancelScheduledMessageWithLogAsync(ServiceBusSender sender, 
             long sequenceNumber, WriteToLogDelegate writeToLog)
         {
-            await sender.CancelScheduledMessageAsync(sequenceNumber);
-            writeToLog($"Cancelled scheduled message with sequence number {sequenceNumber}.");
+            var task = sender.CancelScheduledMessageAsync(sequenceNumber).ContinueWith(
+               (t, state) => writeToLog($"Cancelled scheduled message with sequence number {state}."),
+                sequenceNumber);
+            return task;
         }
     }
 }
--- a/src/ServiceBus/Helpers/CancelScheduledMessagesHelper.cs
+++ b/src/ServiceBus/Helpers/CancelScheduledMessagesHelper.cs
@@ -1,0 +1,32 @@
+﻿using System.Collections.Generic;
+using System.Diagnostics;
+using System.Threading.Tasks;
+
+namespace ServiceBusExplorer.ServiceBus.Helpers
+{
+    public static class CancelScheduledMessagesHelper
+    {
+        public static async Task CancelScheduledMessages(ServiceBusHelper2 serviceBusHelper,
+            string queueName, List<long> sequenceNumbersToCancel)
+        {
+            var client = serviceBusHelper.CreateServiceBusClient();
+            var sender = client.CreateSender(queueName);
+
+            serviceBusHelper.WriteToLog($"Starting cancellation of scheduled messages on queue {queueName}.");
+
+            var stopwatch = Stopwatch.StartNew();
+
+            // Got lots of Azure.Messaging.ServiceBus.ServiceBusExceptions and the UI stopped working when testing 
+            // with 1000 messages when calling using Task.WhenAll. So, doing it one by one instead.
+            foreach (var sequenceNumber in sequenceNumbersToCancel)
+            {
+                await sender.CancelScheduledMessageAsync(sequenceNumber);
+                serviceBusHelper.WriteToLog($"Cancelled scheduled message with sequence number {sequenceNumber}.");
+            }
+
+            stopwatch.Stop();
+
+            serviceBusHelper.WriteToLog($"Cancelled {sequenceNumbersToCancel.Count} scheduled message(s) in {stopwatch.ElapsedMilliseconds / 1000} seconds.");
+        }
+    }
+}
--- a/src/ServiceBus/Helpers/CancelScheduledMessagesHelper.cs
+++ b/src/ServiceBus/Helpers/CancelScheduledMessagesHelper.cs
@@ -30,7 +30,7 @@ namespace ServiceBusExplorer.ServiceBus.Helpers
                 {
                     await semaphore.WaitAsync();
                     tasks.Add(CancelScheduledMessageWithLogAsync(sender, sequenceNumber, serviceBusHelper.WriteToLog)
-                        .ContinueWith((t) => semaphore.Release()));
+                        .ContinueWith((t, state) => ((SemaphoreSlim)state)?.Release()));
                 }
 
                 await Task.WhenAll(tasks);
--- a/src/ServiceBus/Helpers/CancelScheduledMessagesHelper.cs
+++ b/src/ServiceBus/Helpers/CancelScheduledMessagesHelper.cs
@@ -45,15 +45,16 @@ namespace ServiceBusExplorer.ServiceBus.Helpers
                 await Task.WhenAll(tasks);
                 stopwatch.Stop();
 
-                Func<string> successesSingleOrPlural = () => successes.Count > 1 ? "messages" : "message";
+                Func<int, string> singleOrPlural = (c) => c > 1 ? "messages" : "message";
 
-                serviceBusHelper.WriteToLog($"Successfully cancelled {successes.Count} scheduled {successesSingleOrPlural()} in {stopwatch.Elapsed}.");
+
+                serviceBusHelper.WriteToLog($"Successfully cancelled {successes.Count} " +
+                    $"scheduled {singleOrPlural(successes.Count)} in {stopwatch.Elapsed}.");
 
                 if (failures.Count > 0)
                 {
-                    Func<string> failuresSingleOrPlural = () => failures.Count > 1 ? "messages" : "message";
-
-                    serviceBusHelper.WriteToLog($"Failed to cancel {failures.Count} {failuresSingleOrPlural()}.");
+                    serviceBusHelper.WriteToLog($"Failed to cancel {failures.Count} " +
+                        $"{singleOrPlural(failures.Count)}.");
                 }
             }
             finally
--- a/src/ServiceBus/Helpers/CancelScheduledMessagesHelper.cs
+++ b/src/ServiceBus/Helpers/CancelScheduledMessagesHelper.cs
@@ -12,9 +12,10 @@ namespace ServiceBusExplorer.ServiceBus.Helpers
 {
     public static class CancelScheduledMessagesHelper
     {
-        class PassIntToTask
+        class OperationStatus
         {
-            public int Count; 
+            public int Successes;
+            public int Failures;
         }
 
         public static async Task CancelScheduledMessages(ServiceBusHelper2 serviceBusHelper,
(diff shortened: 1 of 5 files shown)


src/ServiceBusExplorer/Controls/HandleQueueControl.cs
```diff
@@ -523,6 +523,21 @@ public void GetMessageSessions()
...
+            {
+                var brokeredMessage = row.DataBoundItem as BrokeredMessage;
+
+                if (brokeredMessage?.State != MessageState.Scheduled)
+                {
+                    return false;
+                }
+            }
+
+            return true;
```
SeanFeldman, line 538: It's not critical, but could be expressed as a linq statement.

src/ServiceBusExplorer/Controls/HandleQueueControl.cs
```diff
@@ -3353,6 +3386,59 @@ private void resubmitSelectedMessagesInBatchModeToolStripMenuItem_Click(object s
             ResubmitSelectedMessages();
         }
 
+
+        async void cancelScheduledMessageToolStripMenuItem_Click(object sender, EventArgs e)
+        {
+            try
+            {
+                if (messagesDataGridView.SelectedRows.Count <= 0)
```
SeanFeldman, line 3394: Doesn't need to be in the try/catch, right?

src/ServiceBusExplorer/Controls/HandleQueueControl.cs
```diff
@@ -3353,6 +3386,59 @@ private void resubmitSelectedMessagesInBatchModeToolStripMenuItem_Click(object s
...
+                    serviceBusHelper.ConnectionString, 
+                    queueDescription.Path);
+                var sequenceNumbersToCancel = messages.Select(s => s?.SequenceNumber).ToList();
+
+                foreach (var sequenceNumber in sequenceNumbersToCancel)
+                {
+                    if (null != sequenceNumber)
+                    {
+                        await queueClient.CancelScheduledMessageAsync((long)sequenceNumber);
+                        writeToLog($"Cancelled scheduled message with sequence number {sequenceNumber}.");
```
SeanFeldman, line 3432: Suggestion:
Aggregate all tasks and use `Task.WhenAll`. Then, log all sequence numbers if that's necessary.

src/ServiceBusExplorer/Controls/HandleQueueControl.cs
```diff
@@ -3374,6 +3460,18 @@ private void ResubmitSelectedMessages()
...
+        void SetCancelScheduledMessageToolStripMenuItemText(bool multipleSelectedRows)
+        {
+            if (multipleSelectedRows)
+            {
+                cancelScheduledMessageToolStripMenuItem.Text = "Cancel Scheduled Messages";
+            }
+            else
+            {
+                cancelScheduledMessageToolStripMenuItem.Text = "Cancel Scheduled Message";
+            }
```
SeanFeldman, line 3472:
```suggestion
             cancelScheduledMessageToolStripMenuItem.Text = multipleSelectedRows
              ? "Cancel Scheduled Messages"
              : "Cancel Scheduled Message";
```

SeanFeldman (changes requested):

SeanFeldman:
Looking good, @ErikMogensen. A few questions/suggestions.

src/ServiceBusExplorer/Controls/HandleQueueControl.cs
```diff
@@ -3353,6 +3386,59 @@ private void resubmitSelectedMessagesInBatchModeToolStripMenuItem_Click(object s
...
+                var messages = messagesDataGridView.SelectedRows.Cast<DataGridViewRow>()
+                    .Select(r => (BrokeredMessage)r.DataBoundItem);
+                var queueClient = QueueClient.CreateFromConnectionString(
+                    serviceBusHelper.ConnectionString, 
+                    queueDescription.Path);
+                var sequenceNumbersToCancel = messages.Select(s => s?.SequenceNumber).ToList();
+
+                foreach (var sequenceNumber in sequenceNumbersToCancel)
+                {
+                    if (null != sequenceNumber)
```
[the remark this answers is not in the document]
ErikMogensen: Otherwise the compiler complains about a null value.

src/ServiceBusExplorer/Controls/HandleQueueControl.cs
```diff
@@ -523,6 +523,21 @@ public void GetMessageSessions()
 
         #region Private Methods
 
+        bool AreAllSelectedMessageScheduled(DataGridViewSelectedRowCollection selectedRows)
```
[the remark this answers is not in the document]
ErikMogensen: Fixed

src/ServiceBusExplorer/Controls/HandleQueueControl.cs
ErikMogensen: Changed

src/ServiceBusExplorer/Controls/HandleQueueControl.cs
ErikMogensen: I'd say that the logging is necessary since it is the only indication you get that if it succeeds. 

If one task fails it will be good to know which messages were cancelled. It is of course possible to create a separate task for doing both operations, but I think that's overkill.

src/ServiceBusExplorer/Controls/HandleQueueControl.cs
ErikMogensen: I assumed it needed that since it is called from the UI thread, but apparently it is handled by the registered exception handler 😄

src/ServiceBusExplorer/Controls/HandleQueueControl.cs
```diff
@@ -3353,6 +3381,52 @@ private void resubmitSelectedMessagesInBatchModeToolStripMenuItem_Click(object s
...
+            var messages = messagesDataGridView.SelectedRows.Cast<DataGridViewRow>()
+                .Select(r => (BrokeredMessage)r.DataBoundItem);
+            var queueClient = QueueClient.CreateFromConnectionString(
+                serviceBusHelper.ConnectionString, 
+                queueDescription.Path);
+            var sequenceNumbersToCancel = messages.Select(s => s?.SequenceNumber).ToList();
+
+            foreach (var sequenceNumber in sequenceNumbersToCancel)
+            {
+                if (null != sequenceNumber)
```
SeanFeldman, line 3422: Something doesn't add up here.
If we have a list of messages, those must have sequence numbers. There's no message w/o a sequence number. I'd double-check that as the null check here feels off.

src/ServiceBusExplorer/Controls/HandleQueueControl.cs
```diff
@@ -3353,6 +3381,52 @@ private void resubmitSelectedMessagesInBatchModeToolStripMenuItem_Click(object s
...
+                {
+                    await queueClient.CancelScheduledMessageAsync((long)sequenceNumber);
```
SeanFeldman, line 3424: When we get the sequence numbers into a list, that's when we can cast to `long`.

src/ServiceBusExplorer/Controls/HandleQueueControl.cs
```diff
@@ -3353,6 +3381,52 @@ private void resubmitSelectedMessagesInBatchModeToolStripMenuItem_Click(object s
...
+                    writeToLog($"Cancelled scheduled message with sequence number {sequenceNumber}.");
```
SeanFeldman, line 3425: We could have a helper method that would take a queue client, sequence number and would cancel the message and log the sequence number. That method could be a task call for each sequence number. That way, you could use `Task.WhenAll`. For scenarios when someone is cancelling a large amount of messages, it will be more helpful than executing sequentially.

SeanFeldman (approved):
I'm not going to block, but I'd like you to double-check a couple of comments/suggestions. Thanks, Erik!

src/ServiceBusExplorer/Controls/HandleQueueControl.cs
ErikMogensen: Yes, that's what I wrote in my comment 😄 

I implemented that as 

`await Task.WhenAll(sequenceNumbersToCancel.Select(s => CancelScheduledMessageWithLogAsync(queueClient, s)));
`

and a new method:

```
async Task CancelScheduledMessageWithLogAsync(QueueClient queueClient, long sequenceNumber)
{
    await queueClient.CancelScheduledMessageAsync(sequenceNumber);
    writeToLog($"Cancelled scheduled message with sequence number {sequenceNumber}.");
}
```

However, that didn't work when cancelling 10 000 messages. The UI froze and it started throwing exceptions. Looking at the implementation for the logging I am pretty sure it is not thread safe. 

So I reverted the code and brought it back to the original.

src/ServiceBusExplorer/Controls/HandleQueueControl.cs
ErikMogensen: I agree. When I removed the null conditional operator in 
`var sequenceNumbersToCancel = messages.Select(s => s?.SequenceNumber).ToList();`
it stopped complaining about null. However, if an item in the list is null it will pass null to the method. I don't see how that could happen though so now it is without that operator.

SeanFeldman: I would suggest filtering out any message that is coming from the grid as `null` and getting rid of the conditional execution.

```
 var sequenceNumbersToCancel = messages.Where(x => x is not null).Select(x => x.SequenceNumber).ToList();
```

src/ServiceBusExplorer/Controls/HandleQueueControl.cs
SeanFeldman: See comment [here](https://github.com/paolosalvatori/ServiceBusExplorer/pull/746/files/e16d80d332c1839011f664e57a08fe88546357eb#r1378086040).

src/ServiceBusExplorer/Controls/HandleQueueControl.cs
SeanFeldman: `WriteToLog` is a delegate passed in from the `MainForm` when the control is constructed ([reference](https://github.com/paolosalvatori/ServiceBusExplorer/blob/3cfbe19e4a1c5e083496b2af905ff33df02f2656/src/ServiceBusExplorer/Forms/MainForm.cs#L4736)).
There's also a [`AsyncWriteToLog`](https://github.com/paolosalvatori/ServiceBusExplorer/blob/3cfbe19e4a1c5e083496b2af905ff33df02f2656/src/ServiceBusExplorer/Forms/MainForm.cs#L3200). Given the context here is asynchronous, should the asynchronous version be used?

src/ServiceBusExplorer/Controls/HandleQueueControl.cs
ErikMogensen: I just saw your comment after implementing a change in a different way.

src/ServiceBusExplorer/Controls/HandleQueueControl.cs
ErikMogensen: I missed that one. I'll take a look at it.

ErikMogensen:
@SeanFeldman, there were quite a few changes since your review. Do you want to take another look?

When I limited it to 40 parallel tasks it is quite fast and works 😄 
image

I also added a new context menu item for selecting all rows

Selected a single, normal messages:
image

Selected multiple, normal messages:
image

Selected a single, dead-letter messages:
image

Selected multiple, deadletter messages:
image

src/ServiceBus/Helpers/ServiceBusHelper2.cs
```diff
@@ -44,6 +62,14 @@ public bool ConnectionStringContainsEntityPath()
             return false;
         }
 
+        // Remember to dispose of the client after use, using client.DisposeAsync()
```
SeanFeldman, line 65: Is this for the callers?

ErikMogensen: Yes. What do you think of changing the comment to: 
`// Callers shall dispose the client after use, using client.DisposeAsync()`

?

I was thinking of creating a public method that only did a client.DisposeAsync() to get symmetry but felt it was overkill.

SeanFeldman: The wording is ok.
It's hard to enforce that with an inline comment. I was thinking that maybe it should be a `<remark>` on the mehothod's XML doc (`<summary>`).

ErikMogensen: I just pushed an update for that comment. Are we good to go?

SeanFeldman: Yep. Just added a tiny suggestion [here](https://github.com/paolosalvatori/ServiceBusExplorer/pull/746/files/522bcd5825093667866c1daa8d45d38d4edba8eb..b8486f154c9ae7720d6e23e77cf2acac6abfd43b#r1490169974).

src/ServiceBus/Helpers/CancelScheduledMessagesHelper.cs
```diff
@@ -0,0 +1,53 @@
...
+
+                var stopwatch = Stopwatch.StartNew();
+                var semaphore = new SemaphoreSlim(40); // As recommended by https://learn.microsoft.com/en-us/azure/service-bus-messaging/message-transfers-locks-settlement#settling-send-operations
+                var tasks = new List<Task>(sequenceNumbersToCancel.Count);
+
+                foreach (long sequenceNumber in sequenceNumbersToCancel)
+                {
+                    await semaphore.WaitAsync();
+                    tasks.Add(CancelScheduledMessageWithLogAsync(sender, sequenceNumber, serviceBusHelper.WriteToLog)
+                        .ContinueWith((t) => semaphore.Release()));
```
SeanFeldman, line 33:
```suggestion
                        .ContinueWith(t => semaphore.Release()));
```

src/ServiceBus/Helpers/CancelScheduledMessagesHelper.cs
```diff
@@ -0,0 +1,53 @@
...
+                }
+
+                await Task.WhenAll(tasks);
+                stopwatch.Stop();
+                serviceBusHelper.WriteToLog($"Cancelled {sequenceNumbersToCancel.Count} scheduled message(s) in {stopwatch.ElapsedMilliseconds / 1000} seconds.");
```
SeanFeldman, line 38: I would go with `Elapsed` because it's a `TimeSpan` formatted automatically. If purging takes a few milliseconds, seconds, or minutes, it will always be in the same readable format `00:00:00.0000000`.

```suggestion
                Func<string> singleOrPlural = () => sequenceNumbersToCancel.Count() > 1 ? "messages": "message";
                serviceBusHelper.WriteToLog($"Cancelled {sequenceNumbersToCancel.Count} scheduled {singleOrPlural()} in {stopwatch.Elapsed}.");
```

src/ServiceBus/Helpers/CancelScheduledMessagesHelper.cs
```diff
@@ -0,0 +1,53 @@
...
+            {
+               await client.DisposeAsync();
+            }
+        }
+
+        static async Task CancelScheduledMessageWithLogAsync(ServiceBusSender sender, 
+            long sequenceNumber, WriteToLogDelegate writeToLog)
+        {
+            await sender.CancelScheduledMessageAsync(sequenceNumber);
+            writeToLog($"Cancelled scheduled message with sequence number {sequenceNumber}.");
```
SeanFeldman, line 50: @ErikMogensen, I don't think you want to wait here... you want to fire and hand over the task to the caller (`CancelScheduledMessages`).
```suggestion
            var task = sender.CancelScheduledMessageAsync(sequenceNumber).ContinueWith(
               (t, state) => writeToLog($"Cancelled scheduled message with sequence number {state}."),
                sequenceNumber);
            return task;
```

SeanFeldman (approved):

SeanFeldman:
Some thoughts about this PR in light of some benchmarking that was done.

It is worth mentioning that this is related to #360. We closed the issue as `won't fix` because, with any significant number of scheduled messages, this approach (PeekLock and delete) might fail or take a long time. For example, running outside of an Azure data center and cancelling 500K messages of average size can take anywhere between 1 and 1.5 hours.

I also benchmarked an alternative approach, which is more intrusive, using auto-forwarding and a temporary queue. Assuming there's a `source` queue with scheduled and active messages, auto-forwarding is set to a temporary queue `temp`. Active messages are drained to the temporary `temp` queue. At that point, the original `source` queue is deleted and recreated (a short delay is required to avoid resource creation conflict error). And the last step is to forward backed-up messages from the `temp` queue back to the newly re-created `source`, deleting the temporary `temp` queue. With a similar number of scheduled messages and 1000 active messages to emulate in-flight messages, it took under 30 seconds (manual verification).

While I still rather see ASB [supporting purging](https://github.com/Azure/azure-service-bus/issues/1) (scenario when there are no inflight messages or those don't matter) and ability to get [peek messages by status](https://github.com/Azure/azure-service-bus/issues/95), for large quantities of scheduled messages, the auto-forwarding option might be better suited. Foot for thought

ErikMogensen:
Fixed it to build and handle failed cancellations. Tested with 100.000 messages and it took a bit less than ten minutes to cancel all of them.

@SeanFeldman, interesting solution. What do you mean by:
> this approach (PeekLock and delete)

SeanFeldman:
I meant PeekLock and complete.
The alternative is too destructive and would mess up anything that relies on the resource ID. Like monitoring and alerting. That’s why it really should be a service side capability (https://github.com/Azure/azure-service-bus/issues/698).

src/ServiceBus/Helpers/CancelScheduledMessagesHelper.cs
```diff
@@ -22,34 +28,57 @@ public static class CancelScheduledMessagesHelper
...
-                Func<string> singleOrPlural = () => sequenceNumbersToCancel.Count() > 1 ? "messages": "message";
-                serviceBusHelper.WriteToLog($"Cancelled {sequenceNumbersToCancel.Count} scheduled {singleOrPlural()} in {stopwatch.Elapsed}.");
+
+                Func<string> successesSingleOrPlural = () => successes.Count > 1 ? "messages" : "message";
+
+                serviceBusHelper.WriteToLog($"Successfully cancelled {successes.Count} scheduled {successesSingleOrPlural()} in {stopwatch.Elapsed}.");
+
+                if (failures.Count > 0)
+                {
+                    Func<string> failuresSingleOrPlural = () => failures.Count > 1 ? "messages" : "message";
```
SeanFeldman, line 54: It’s identical to `successesSingleOrPlural`. Merge the two?

src/ServiceBus/Helpers/CancelScheduledMessagesHelper.cs
```diff
@@ -22,34 +28,57 @@ public static class CancelScheduledMessagesHelper
...
                 var stopwatch = Stopwatch.StartNew();
                 var semaphore = new SemaphoreSlim(40); // As recommended by https://learn.microsoft.com/en-us/azure/service-bus-messaging/message-transfers-locks-settlement#settling-send-operations
                 var tasks = new List<Task>(sequenceNumbersToCancel.Count);
 
                 foreach (long sequenceNumber in sequenceNumbersToCancel)
                 {
                     await semaphore.WaitAsync();
-                    tasks.Add(CancelScheduledMessageWithLogAsync(sender, sequenceNumber, serviceBusHelper.WriteToLog)
-                        .ContinueWith((t, state) => ((SemaphoreSlim)state)?.Release()));
+                    tasks.Add(CancelScheduledMessageWithLog(sender, sequenceNumber, serviceBusHelper.WriteToLog, successes, failures)
```
SeanFeldman, line 41: What do you think of combining the success and the failure counters into a `OperationStatus`?

ErikMogensen: You mean create a class like this?

```
class OperationStatus
{
    public int Successes; 
    public int Failures; 
}
```

SeanFeldman: Yeah, something like that. Those two counters represent a result of a single operation and are coupled.

ErikMogensen: That's right. Pushed a change to use such a class.

ErikMogensen:
> I meant PeekLock and complete.
Do you mean the solution I did and you just reviewed? It is using Peek to find out status and then CancelScheduledMessages on the scheduled messages.

SeanFeldman:
> Do you mean the solution I created and you just reviewed? It is using Peek to find out status and then CancelScheduledMessages on the scheduled messages.

Yes. There's no other option today. Except the one I mentioned but it's not "safe".

ErikMogensen:
@SeanFeldman, do you approve of this PR now?

SeanFeldman:
> @SeanFeldman, do you approve of this PR now?

image

👍

PR #747, opened 2023-11-07
Change the way the tenantId is used in the Event Grid login

ErikMogensen:
The tenantId field was always null when calling InteractiveBrowserCredential. This seems to work with work accounts but not personal Microsoft accounts. 

This PR sets the tenantId to the custom tenantId, if it is specified.

### DIFF
--- a/src/EventGridExplorerLibrary/EventGridLibrary.cs
+++ b/src/EventGridExplorerLibrary/EventGridLibrary.cs
@@ -36,19 +36,18 @@ namespace EventGridExplorerLibrary
         private EventGridManagementClient controlPlaneClient;
         private Dictionary<string, EventGridClient> dataPlaneClients = new Dictionary<string, EventGridClient>();
         private int retryTimeout;
-        private string tenantId;
         private readonly WriteToLogDelegate writeToLog = default;
         #endregion
 
         public EventGridLibrary(string subscriptionId, string apiVersion, int retryTimeout, string cloudTenant, string customId, WriteToLogDelegate writeToLog)
         {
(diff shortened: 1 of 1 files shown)


src/EventGridExplorerLibrary/EventGridLibrary.cs
```diff
@@ -37,20 +37,19 @@ public class EventGridLibrary
...
         #endregion
 
         public EventGridLibrary(string subscriptionId, string apiVersion, int retryTimeout, string cloudTenant, string customId, WriteToLogDelegate writeToLog)
         {
+            string tenantId = customId == string.Empty ? null : customId;
             var apiVersionToUse = string.IsNullOrEmpty(apiVersion) ? DefaultApiVersion : apiVersion;
-            controlPlaneClient = new EventGridManagementClient(new Uri(ArmEndpointUrl), GetTokenCredential(), subscriptionId, apiVersionToUse , retryTimeout);
+            controlPlaneClient = new EventGridManagementClient(new Uri(ArmEndpointUrl), GetTokenCredential(tenantId), subscriptionId, apiVersionToUse , retryTimeout);
             this.retryTimeout = retryTimeout;
-            this.tenantId = customId == string.Empty ? tenantIds[cloudTenant] : customId;
```
SeanFeldman, line 49: If the passed in `tenantId` is used, what's `tenantIds`array is used for?

ErikMogensen: I don't know what the original intent was. Hopefully @hillaryc can answer that. I have no way of testing any other cloud than Public. I think we should replace the cloud code names with the actual names even though the code names are cool 😎 

Mooncake = China Cloud
Fairfax = US Government Cloud

Sean, have you tested connecting to event grids in your subscriptions with this PR?

paolosalvatori (approved):

Issue #607, opened 2021-11-29
Windows AAD login


ChoOo7:
Hello

Can we have an Azure Active directory login to have access to our services bus ?

We don't want to have any password / keys on our computers

Thanks

ErikMogensen:
@ChoOo7, that's something that's on the roadmap. However, currently 95% of the code Service Bus Explorer is using to communicate with queues and topics is using the oldest Service Bus SDK. The remaining 5% is using the latest Service Bus SDK. For several reasons we want to migrate everything to the latest SDK and that has a higher priority for me. 

However, you or anyone else is most welcome to come up with a design suggestion for AAD login. I recommend starting with that before creating a PR since this a complicated issue.

Rookian:
I think the easiest solution is to use `AzureCliCredential` or `InteractiveBrowserCredential` when authenticating against Azure Service Bus. The new SDK already provides an option to use `TokenCredential`.
```csharp
 public ServiceBusClient(string fullyQualifiedNamespace, TokenCredential credential);
 ```

SeanFeldman:
@Rookian 
>  However, currently 95% of the code Service Bus Explorer is using to communicate with queues and topics is using the oldest Service Bus SDK.

kensykora:
I really want this feature so we can stop using access keys!

jwerner77:
Azure AD auth will be a huge win for this tool!  There is some very valuable functionality such as fix/resubmit from dead letter queue which are not available in the portal.  However the only way for a team to use this is they all have to have connection string with the keys which is not secure.

rangp:
Is there any update on this? 
I just looked at the code and it seems as if the old sdk has been removed. With the new sdk it should be fairly simple to integrate this, since both the ServiceBusClient and ServiceBusAdministrationClient support passing a TokenCredential. 

I can see two ways of integrating this:

1. Add a ui feature that allows one to specify whether to authenticate via active directory or not. Something like this. And if specified, a flag is passed that allows the ServiceBusClient and ServiceBusAdministrationClient to be constructed with InteractiveBrowserCredentials. This would then open a browser window where one can sign in and the rest should be handled by the sdk out of the box
![mockup](https://user-images.githubusercontent.com/8915141/165271679-a2584d06-70e8-431d-b96c-dd2bad675c75.png)

[...]

PR #750, opened 2023-11-13
Sdk migration ServiceBus Queues

Merlijnv:
To make #607 possible we need to migrate the sdk's for servicebus, notificationhub and such to the new sdk.
To make this possible I am going to start with servicebus queues


Merlijnv:
@SeanFeldman the PR you asked for so that the issue is not filled with discussions about the implementatie and such

Merlijnv:
@SeanFeldman previous QueueDescription seems to be splitted up into 2 (QueueProperties and QueueRuntimeProperties) So I have to or choose 1 to use or retrieve both. What is your opinion on this?

SeanFeldman:
> @SeanFeldman previous QueueDescription seems to be splitted up into 2 (QueueProperties and QueueRuntimeProperties) So I have to or choose 1 to use or retrieve both. What is your opinion on this?

Not following.
- `QueueProperties` is necessary to provide the information about the queue and to be able to modify those.
- `QueueRuntimeProperties` is for the message counts.

You'll need both. If you want, you could introduce an abstraction, `QueueProperties` that would expose the information from the two new APIs.

Merlijnv:
> Not following.
> 
>     * `QueueProperties` is necessary to provide the information about the queue and to be able to modify those.
> 
>     * `QueueRuntimeProperties` is for the message counts.

Yeah but for example for the queue overview I only need `QueueRuntimeProperties` because that includes all the info for it.
So I was thinking or always combine and have some extra info but also extra possible unneeded call. Or make seperate calls for if both, `Properties` or `RuntimeProperties` are needed. Idk what would be the impact of always using a combined call.

Frulfump:
> To make #607 possible we need to migrate the sdk's for servicebus, notificationhub and such to the new sdk. To make this possible I am going to start with servicebus queues

#607 is closed as completed so this wasn't necessary for that, but would be good for other reasons

PR #751, opened 2023-11-17
Updated NuGet packages including WindowsAzure.ServiceBus

ErikMogensen:
Also removed unused references.


SeanFeldman (approved):
We should set up Dependably for these.

PR #752, opened 2023-11-19
Improve UI event grid explorer

ErikMogensen:
Changed the menus on the Mainform:
image

On the Connect form I changed the order and width of the fields. I also made the API version a combo box. I did not the set list of tenants associated with the logged in account since I am pretty sure you need to log in the user first and that would be out of scope for this PR. However, I think that is the way to go.

I did not change the selection of the Cloud tenant, since I am not sure how that should work. I set the icon and made the OK button the Accept button and the Cancel button the Escape button. 

image

For the three other dialogs I set the window position to CenterParent, window frame to Fixed and set Accept and Escape buttons:
image

image

image


src/ServiceBusExplorer/Forms/EventGridConnectForm.cs
```diff
@@ -34,8 +34,10 @@ private void btnOk_Click(object sender, EventArgs e)
             ResourceGroup = txtResourceGroupName.Text.Trim();
             NamespaceName = txtNamespaceName.Text.Trim();
             SubscriptionId = txtSubscriptionId.Text.Trim();
-            ApiVersion = txtApiVersion.Text.Trim();
-            RetryTimeout = txtRetryTimeout.Text != string.Empty ? int.Parse(txtRetryTimeout.Text) : 0;
+            ApiVersion = cboApiVersion.Text.Trim(); 
+
+            var retryTimoutInSeconds = txtRetryTimeout.Text != string.Empty ? int.Parse(txtRetryTimeout.Text) : 0;
```
SeanFeldman, line 39:
```suggestion
            int.TryParse(txtRetryTimeout.Text, out var retryTimoutInSeconds);
```

SeanFeldman (approved):
Added one suggestion.
Looks good.

PR #761, opened 2024-01-01
Put back the submenus for Set Status

### LINKED ISSUE #760: The Set Status context menu does not work
The Set Status context menu does not work anymore.  It is not showing the statuses, see figure below.
image

ErikMogensen:
It now looks like this again:
image

Seems that these changes were lost when another PR was merged.

Fixes #760


SeanFeldman (approved):
And that's, ids, why tests are important 😀
Thank you, @ErikMogensen.

ErikMogensen:
> And that's, ids, why tests are important 😀

@SeanFeldman, What's "ids"? 

I agree, more unit tests would be great.

SeanFeldman:
kids 😂

Issue #633, opened 2022-04-07
Resubmit Selected Messages In Batch Mode issue


BruAlcaraz:
Hi, I am facing an issue sometimes when I use the Batch Mode to resubmit multiple messages. 
Sometimes the messages resubmitted have a string added to it '@strin3http://schemas.microsoft.com/2003/10/Serialization/�'
It has happened sometimes not every time. Can anyone help me with that?

![image](https://user-images.githubusercontent.com/20819845/162193849-a487555d-95db-4335-b581-4f0fcf803392.png)

Current Version that I am using is 5.0.7 but it happened on other versions as well

ErikMogensen:
Strange. 

1. It may be the way the messages are displayed. I think the tool guesses how the message should be displayed. Please check a message that looks **good** in Service Bus Explorer in another tool, such as the Azure Portal Service Bus Explorer, if those characters exist in that message. 

2. If those characters are not there in question 1 I wonder if you are sure they are added by Service Bus Explorer?

3. Does this happen if you resubmit messages one by one?

SeanFeldman:
@BruAlcaraz, the sender of the message is using a really old SDK and likely sending the payload which gets serialized with old ASB's internal serializer rather than a stream of bytes. What's the SDK used to generate the message and the specific code that sends it?

BruAlcaraz:
Hi @ErikMogensen 
1) in the Original message the characters are not there.
2) they only get there after I resend via batch but strangely enough the issue does not happen always.
3) no it never happens when resending one by one. But once the issue happens in that batch that I need to resend one by one and make sure to delete the added characters.

BruAlcaraz:
> @BruAlcaraz, the sender of the message is using a really old SDK and likely sending the payload which gets serialized with old ASB's internal serializer rather than a stream of bytes. What's the SDK used to generate the message and the specific code that sends it?

I'll double check it in the morning and get back to you. But as I mentioned to Erik the characters only shows when the messages are resent in a batch never happened on any originally sent messages.

SeanFeldman:
> the characters only shows when the messages are resent in a batch never happened on any originally sent messages.

That's good, let's narrow the scope to that code then, to see how the message batch is resubmitted. Thanks.

OwainRichardson:
Has there been any updates or finds on this thread. It's an issues I'm seeing also when re-submitting in batch.

FredrikDahlberg:
[...]

PR #764, opened 2024-01-09
Fix Issue #633: Resubmit Selected Messages In Batch Mode

Anshuljkt:
To address issue #633 (Resubmit Selected Messages In Batch Mode issue) 

I have added a check on the messageType's value and parsed it as a string if it is one. Also removed the added serialization text `"@\u0006string\b3http://schemas.microsoft.com/2003/10/Serialization/�G"` to prevent messages from becoming corrupted when resubmitting.

Using this build, you should be able to resubmit batches of messages without running into message body corruption.

### DIFF
--- a/src/ServiceBusExplorer/Forms/MessageForm.cs
+++ b/src/ServiceBusExplorer/Forms/MessageForm.cs
@@ -412,18 +412,21 @@ namespace ServiceBusExplorer.Forms
                                     var messageText = serviceBusHelper.GetMessageText(message,
                                         MainForm.SingletonMainForm.UseAscii, out bodyType);
 
-                                    if (messageText.GetType() == typeof(string))
+                                    if (bodyType == BodyType.ByteArray)
+                                    {
+                                        // For body type ByteArray cloning is not an option. When cloned, supplied body can be only of a string or stream types, but not byte array :(
+                                        outboundMessage = message.CloneWithByteArrayBodyType(messageText, messagesSplitContainer.Visible);
+                                    }
+                                    else if (messageText.Contains("@\u0006string\b3http://schemas.microsoft.com/2003/10/Serialization/�G"))
                                     {
                                         // Remove any serialization text from the message body string
                                         messageText = messageText.Replace("@\u0006string\b3http://schemas.microsoft.com/2003/10/Serialization/�G", string.Empty);
                                         outboundMessage = message.Clone(messageText, messagesSplitContainer.Visible);
-                                    }
+                                    } 
                                     else
                                     {
-                                        // For body type ByteArray cloning is not an option. When cloned, supplied body can be only of a string or stream types, but not byte array :(
-                                        outboundMessage = bodyType == BodyType.ByteArray ?
-                                                          message.CloneWithByteArrayBodyType(messageText, messagesSplitContainer.Visible) :
-                                                          message.Clone(message.GetBody<Stream>(), messagesSplitContainer.Visible);
+                                        // Process as Stream
+                                        outboundMessage = message.Clone(message.GetBody<Stream>(), messagesSplitContainer.Visible);
                                     }
                                 }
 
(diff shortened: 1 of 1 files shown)


src/ServiceBusExplorer/Forms/MessageForm.cs
```diff
@@ -412,10 +412,19 @@ private async void btnSubmit_Click(object sender, EventArgs e)
                                     var messageText = serviceBusHelper.GetMessageText(message,
                                         MainForm.SingletonMainForm.UseAscii, out bodyType);
 
-                                    // For body type ByteArray cloning is not an option. When cloned, supplied body can be only of a string or stream types, but not byte array :(
-                                    outboundMessage = bodyType == BodyType.ByteArray ?
-                                                      message.CloneWithByteArrayBodyType(messageText, messagesSplitContainer.Visible) :
-                                                      message.Clone(message.GetBody<Stream>(), messagesSplitContainer.Visible);
+                                    if (messageText.GetType() == typeof(string))
```
ErikMogensen, line 415: I suspect this will always be true since GetMessageText returns a string. Have you tested with the other body types?

SeanFeldman: I second that. The issue here is serialization/deserialization with the legacy SDK (track 0).

Anshuljkt: Thank you for your review! I did not, since we were only dealing with string inputs. I will debug further and see what I can do.

ErikMogensen: Thank you @Anshuljkt!

Anshuljkt: I added back the ByteArray check, but have a question: what is the downside of using the messageText directly for the clone, instead of using the GetBody method?

ErikMogensen: messageText only handles strings whereas GetBody may handle any type.

BodyType is an enum looking like this:

    public enum BodyType
    {
        Stream,
        String,
        Wcf,
        ByteArray
    }

You should probably limit the test to BodyType.String. 

Also, in the unlikely case someone has that long string in their message they will be surprised if it is removed so I think you should only replace it if it is at the first position in the message body. (I assume it is at the first position).

Anshuljkt: The issue I am running into is that the malformed message body (sample below) is classified as BodyType.Stream by the GetMessageText() method, even when the resubmission is done for type "String". So I am not able to check for just BodyType.String here.

```
@�strin3http://schemas.microsoft.com/2003/10/Serialization/�{
  "Prop1": "00000000-0000-0000-0000-000000000000",
  "Prop2": false,
  "Prop3": null,
  "Prop4": "redacted"
}
```

Anshuljkt: I thought of a slightly different solution: avoid overriding the bodyType using the GetMessageText method. This way, the user selection in the resubmission dialog will apply. What do you think about this?

ErikMogensen: > I thought of a slightly different solution: avoid overriding the bodyType using the GetMessageText method. This way, the user selection in the resubmission dialog will apply. What do you think about this?

It would require you to set that for every message, right? I think that people expect bodyType to be automatically set. 

When is that serialization text created?

Anshuljkt: In this scenario, we would set the bodyType for all messages in the batch based on the user input from the dialog box. If the expectation is that the user input is overridden, this solution would not work.

I am not sure where the serialization text comes from...

ErikMogensen: Yes, I am. I suggest the way to fix this is to avoid getting that string. It is probably due to reading a string message using GetBody<Stream>().

ErikMogensen: @Anshuljkt, I tested that and it worked so I created a PR for that, #776.

Anshuljkt:
Thanks for your help, @ErikMogensen! Closing this PR as you have addressed the issue.

PR #765, opened 2024-01-09
Set NUnit.ConsoleRunner to latest version, 3.16.3.

ErikMogensen:
Hopefully this will make the unit tests work again...


SeanFeldman (commented):
I might be wrong, but I think rulesets are deprecated.
Analyzers are now enabled/disabled with `.editorconfig`.

ErikMogensen:
I agree about the deprecation. I got multiple warnings when creating it 😄

ErikMogensen:
@SeanFeldman , I believe you missed that I deleted the ruleset file.

SeanFeldman (approved):

PR #768, opened 2024-02-11
Add selective delete to Transfer deadletter subqueues and fix the context menus for the Transfer Deadletter grid view

ErikMogensen:
Before this PR the context menu when selecting a single Transfer DLQ message looks like this:
image

and when selecting multiple messages it looks like this:
image

This PR removes the separate handling of the Transfer Deadletter context menus, they are now the same as for the Deadletter context menus. A lot of the code for handling messages in these subqueues is now shared. 

Also added the feature to selectively delete Transfer deadletter messages.

The above is only added for queues and not for subscriptions as the transfer deadletter support is a bit mysterious for subscriptions.

### DIFF
--- a/src/ServiceBusExplorer/Forms/MessageForm.cs
+++ b/src/ServiceBusExplorer/Forms/MessageForm.cs
@@ -82,7 +82,7 @@ namespace ServiceBusExplorer.Forms
         #endregion
 
         #region Private Instance Fields
-        SubqueueType subqueueType = SubqueueType.NotSet;
+        QueueType queueType = QueueType.NotSet;
 
         readonly IEnumerable<BrokeredMessage> brokeredMessages;
         readonly BrokeredMessage brokeredMessage;
@@ -118,10 +118,10 @@ namespace ServiceBusExplorer.Forms
         #endregion
 
         #region Public Enums
-        public enum SubqueueType
+        public enum QueueType
         {
             NotSet,
-            NotASubqueue,
+            PrimaryQueue,
             Deadletter,
             TransferDeadletter
         }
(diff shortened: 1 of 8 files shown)


src/ServiceBusExplorer/Forms/MessageForm.cs
```diff
@@ -115,6 +117,16 @@ public partial class MessageForm : Form
         };
         #endregion
 
+        #region Public Enums
+        public enum SubqueueType
+        {
+            NotSet,
+            NotASubqueue,
```
sean-feldman-sh, line 124: PrimaryQueue?

src/ServiceBusExplorer/Forms/MessageForm.cs
```diff
@@ -115,6 +117,16 @@ public partial class MessageForm : Form
         };
         #endregion
 
+        #region Public Enums
+        public enum SubqueueType
```
sean-feldman-sh, line 121: QueueType?

src/ServiceBusExplorer/Forms/MessageForm.cs
```diff
@@ -482,15 +496,20 @@ private async void btnSubmit_Click(object sender, EventArgs e)
...
                         stopwatch.Start();
 
                         if (chkRemove.Checked)
                         {
                             var messageHandler = CreateDeadLetterMessageHandler();
 
                             var result = await messageHandler.MoveMessages(messageSender,
-                                sequenceNumbers, outboundMessages);
+                                sequenceNumbers, 
+                                transferDLQ: subqueueType == SubqueueType.TransferDeadletter ? true: false,
```
sean-feldman-sh, line 508: Ternary condition not necessary here.

sean-feldman-sh (approved):
A few comments. Something is not sitting right with me about `SubqueueType` but not a blocker.

src/ServiceBusExplorer/Forms/MessageForm.cs
ErikMogensen: Sure

src/ServiceBusExplorer/Forms/MessageForm.cs
ErikMogensen: Yes

src/ServiceBusExplorer/Forms/MessageForm.cs
ErikMogensen: Good

SeanFeldman (approved):
Approving with correct account 😄

ErikMogensen:
Approved twice, this must be a fantastic PR 😆

SeanFeldman:
> Approved twice, this must be a fantastic PR 😆

If I wouldn't approve with the correct account, you'd be blocked 😉

ErikMogensen:
> If I wouldn't approve with the correct account, you'd be blocked 😉

I know. I just couldn't resist 😄

PR #770, opened 2024-02-25
Show error message if a config file is incorrect

### LINKED ISSUE #769: If a config file is incorrect the application silently dies
If a config file is incorrect the application silently dies when starting it.

ErikMogensen:
Fixes #769 

Shows one or two error messages if a config file is incorrect. The application still dies after that.

### DIFF
--- a/src/ServiceBusExplorer/Forms/MainForm.cs
+++ b/src/ServiceBusExplorer/Forms/MainForm.cs
@@ -4041,7 +4041,14 @@ namespace ServiceBusExplorer.Forms
         #region Public Static Methods
         public static void StaticWriteToLog(string message, bool async = true)
         {
-            mainSingletonMainForm.WriteToLog(message);
+            if(null != mainSingletonMainForm)
+            {
+                mainSingletonMainForm.WriteToLog(message);
+            }
+            else
+            {
+                MessageBox.Show(message);
+            }
         }
         #endregion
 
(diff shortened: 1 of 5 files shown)


src/ServiceBusExplorer/Forms/MainForm.cs
```diff
@@ -4041,7 +4041,14 @@ private void SetTitle(string prefix)
         #region Public Static Methods
         public static void StaticWriteToLog(string message, bool async = true)
         {
-            mainSingletonMainForm.WriteToLog(message);
+            if(null != mainSingletonMainForm)
```
SeanFeldman, line 4044: Always found this notation awkward to read.
If you’re concerned about accidental null assignment, there’s a more readable way.

```suggestion
            if(mainSingletonMainForm is not null)
```

SeanFeldman (approved):

src/ServiceBusExplorer/Forms/MainForm.cs
SeanFeldman: Will need to revert. Or bump C# language version.

ErikMogensen: Sadly we are stuck on 7.3 as long as we are using .Net Framework. https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version.

PR #779, opened 2024-06-03
Warnings as errors and upgrade setupMsBuild

ErikMogensen:
Try /WarnAsError to block builds with warnings. 
Upgrade SetupMSBuild to Node.js 20 version

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -35,7 +35,7 @@ jobs:
     - uses: actions/checkout@v3
 
     - name: Add MSBuild for VS 2022 to PATH 
-      uses: microsoft/setup-msbuild@v1.1
+      uses: microsoft/setup-msbuild@v2
       with:
         vs-version: '[17.2,18.0)'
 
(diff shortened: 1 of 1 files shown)


.github/workflows/build-test.yml
```diff
@@ -73,7 +73,7 @@ jobs:
 
     - name: Build
       run: |
-        msbuild /m /property:Configuration=${{env.BUILD_CONFIGURATION}},FileVersion=${{env.NUMBER_VERSION}},InformationalVersion=${{inputs.release-version}} ${{env.SOLUTION_FILE_PATH}}
+        msbuild /WarnAsError /m /property:Configuration=${{env.BUILD_CONFIGURATION}},FileVersion=${{env.NUMBER_VERSION}},InformationalVersion=${{inputs.release-version}} ${{env.SOLUTION_FILE_PATH}}
```
SeanFeldman, line 76: It would be better to do that in the solution/projects. That way, it's consistent locally and on the build server. Otherwise, contributors will not see those errors until push their changes up.

SeanFeldman (changes requested):

asos-gurpreetsingh:
I think this would not be needed anymore since the build is now using dotnet build. unless those warning are still a problem in which case those warning need to be tackled.

ErikMogensen:
We do want to block the build when there are errors. Created PR #825 for this.

Issue #758, opened 2023-12-18
Mass transfer messages from dead letter queue?


DerekSMorin:
This might be a question or possibly a feature request.  If we have a few thousand messages in the deadletter queue, I'm wondering what the best way is to transfer all of them back for reprocessing?

Currently I do something like this:
 
![image](https://github.com/paolosalvatori/ServiceBusExplorer/assets/7568856/bc5768d8-1f90-408d-82ee-17a79004bca0)

But if we try more than a few hundred messages we frequently get this:

<06:52:05> Not all the selected messages were moved. This was caused by the time to go through the messages were longer than Lock duration for the queue. Either delete some messages in the dead letter subqueue or increase the Lock duration for the queue.          

           The Lock duration for the queue is currently 60 seconds.
           
I realize altering the lock duration could make a difference, but I'm wondering what the recommended approach is?

I generally end up doing 100-200 messages at a time but this is manual and time-consuming.

ErikMogensen:
I see two alternatives. 

Either:
Set Lock duration to maximum unless you suspect that could cause trouble for clients reading messages from the queue. Maximum Lock duration is five minutes. You may still not be able to  send all messages in the same batch. When done set Lock duration back to what it was.

Or:
Uncheck Remove message from DLQ and send them all. When done, Purge the Dead Letter Queue (DLQ). Before purging, check that no new messages have arrived in the DLQ. There is a risk that if a message gets moved to the DLQ it would be lost though. You can disable Send to the queue when doing this if you are sure that the sending clients can handle that.

SeanFeldman:
> You can disable Send to the queue when doing this if you are sure that the sending clients can handle that.
Risky. Especially when there's no control over senders.

I don't remember the logic to move messages out of DLQ, but it might explain the issue if the code is greedy. Then, the max lock duration on the queue won't help.

PR #780, opened 2024-06-03
#758

kjakee:
Increased the timeout so that it can process a large volume of message from dead letter queue in a batch. Also added a log information so that it logs the sequence number of the message moved from dead letter queue. I hope it fix the issue #758.


src/Common/Helpers/DeadLetterMessageHandler.cs
```diff
@@ -246,7 +246,7 @@ await messageSender.SendAsync(messagesToSend[index])
                                 await message.AbandonAsync().ConfigureAwait(false);
                                 throw;
                             }
-
+                            WriteToLog($"Moving a message with sequence number {message.SequenceNumber}.");
```
SeanFeldman, line 249: Would it make sense to indicate moving from where to where? I.e. 
> Moving message with sequence number 1234567890 from queue.dead-letter to queue

src/Common/Helpers/DeadLetterMessageHandler.cs
```diff
@@ -265,7 +265,7 @@ await messageSender.SendAsync(messagesToSend[index])
                         done = true;
                     }
 
-                    if (stopwatch.ElapsedMilliseconds >= maxTimeInSeconds * 1000)
+                    if (stopwatch.ElapsedMilliseconds >= maxTimeInSeconds * 100000)// changed from 1000 to 100000
```
SeanFeldman, line 268: Please elaborate.
This `maTimeInSeconds` is driven by `GetMaxOperationTimeInSeconds()` which [returns](https://github.com/paolosalvatori/ServiceBusExplorer/blob/18c52d79731775e8e8a845ee6cb75b541d25e0c3/src/Common/Helpers/DeadLetterMessageHandler.cs#L347) `MaxLockDuration` of the entity and that cannot be changed (not today)

SeanFeldman (changes requested):

src/Common/Helpers/DeadLetterMessageHandler.cs
kjakee: Per my understand stopwatch.ElapsedMillisecondsis here is the time of the max lock duration - for a large batch of messages i.e. 2000 messages, the processing time exceeds it if we do not increase it sufficiently. For that reason, I have multiplied maxTimeInSeconds to a large number instead of 1000. The alternate way of doing this might be increasing maxTimeInSeconds instead of increasing 1000 to 100000 here.

src/Common/Helpers/DeadLetterMessageHandler.cs
kjakee: yes, that makes better sense.

.github/workflows/dotnet-desktop.yml
ErikMogensen, line 1: Is this file supposed to be part of the PR?

src/Common/Helpers/DeadLetterMessageHandler.cs
ErikMogensen: If you increase the duration longer than the maximum lock duration, then the messages will most likely not be deleted, which is undesirable.

kjakee: Based on my most recent testing - I tried several times with almost 2000 messages in batch. Every time the messages got cleared up as expected.

ErikMogensen: How long did it take and what is the Max Lock Duration for the queue/subscription?

kjakee: For 2000 messages it took almost 2 and half an hour to reprocess and cleared up the dead letter queue. The max lock duration for the subscription was 1 minute.

ErikMogensen: Hmm, interesting. That's an extremely long time for handling 2,000 messages though. Are you aware of any reason for this?

kjakee: Those 2000 message were sent to reprocess in a subscription which triggers a service bus trigger azure function. I suspect a potential reason was the batch mode was not enabled in that azure function.

PR #788, opened 2024-06-29
Modify Mainform when running VS in DPI unaware mode

ErikMogensen:


SeanFeldman (approved):

deemuzi:
It is not related to the DPI issue but when I try to open ServiceBusExplorer.exe from 6.0.1 release it says "Unknown Publisher". Please see the image below:
image

sean-feldman-sh:
> It is not related to the DPI issue

Then let's not tag it here, @deemuzi.

For the context, the warning that you see is because the executable is not signed.

Issue #784, opened 2024-06-25
UI doesn't scale correctly


Splaxi:
Using the latest release - on Windows 11, Microsoft Surface, the UI doesn't scale:

![image](https://github.com/paolosalvatori/ServiceBusExplorer/assets/8750181/b919d061-53bc-43ac-89d4-188336e51055)

It seems that the right anchor isn't is configured correctly. The scroll bar doesn't show if I minimize to a non-full screen mode.

Is this by design?

maf-soft:
Same here, and after "Set Default Layout", the Log area is completely out of the window and not reachable anymore.

ErikMogensen:
No, that's not the way it is supposed to look.

ErikMogensen:
@Splaxi, @maf-soft I am unable to reproduce this.

1. Are you using v6.0.0?
2. Is it the same in 5.0.18?
3. @splaxi, what is the Scale setting? It is under Settings | System > Display
![image](https://github.com/paolosalvatori/ServiceBusExplorer/assets/9644248/3d76fab3-8472-4680-9df1-8d6f9bf32865)
4. Anything else you can think of that would cause this bahavior?

maf-soft:
Yes, I used choco to update from 5.0.18 (where it looked ok) to 6.0.0. It runs on a VM with WindowsServer 2022-datacenter-azure-edition, the RDP client is Win11.

maf-soft:
Ah the client has multiple screens in my case. If I move the RDP window to the other screen, which has a scaling of 250%, the wrong proportions look exactly the same, just everything 250% larger.

Splaxi:
Single monitor, zip file download directly from GitHub.

Not around my machine at the moment.

No issues in v5 - version was not the latest v5 though....

Splaxi:
Okay.

Microsoft Surface Laptop 4, running directly on battery. No cables connected.

With scale set to 100% - it looks like this:

![image](https://github.com/paolosalvatori/ServiceBusExplorer/assets/8750181/11a0acc1-78d0-4bd7-90b5-e378cfce8a0b)
- All edges and boxes looks like expected

Settings are like this: 
![image](https://github.com/paolosalvatori/ServiceBusExplorer/assets/8750181/9485c97d-3b1e-421e-b1e0-3ee94cbb597c)

Anything other than 100%, eg. 125/150% - yields the following result:

[...]

PR #790, opened 2024-07-04
Use new functionality to force VS run in DPI unaware mode

ErikMogensen:
According to https://learn.microsoft.com/en-us/dotnet/desktop/winforms/whats-new/net80?view=netdesktop-8.0#visual-studio-dpi-improvements this will force VS to load the project in the selected DPI awareness mode. Unfortunately it does not work on my machine, but it may work on someone else's. 

I think anything that lessens the risk of the UI scaling issue #784 happening again is worth a lot.


SeanFeldman (approved):
Let's hope it does the trick 🤞

PR #797, opened 2024-07-28
First set changes to enable high DPI in Service Bus Explorer

jmoralesv:
* Adding support for .NET Framework 4.7.2 and using the settings in the app.manifest file for enabling high DPI support, and for taking advantage of some new helper methods the framework offers for high DPI related calculations, such as LogicalToDeviceUnits and ScaleBitmapLogicalToDevice
* Changes for making the ConnectForm and MainForm work with high DPI, by fixing how the height of some controls are calculated, such as HeaderPanel or HandleTopicControl.
* Whitespace formatting in the FilterForm.
* Fixes for the message dialogs ChangeStatusForm, ClipboardForm and DeleteForm, so the message text is displayed completely. 
* Changes in the ParameterForm and the MainForm to display the ParameterForm properly in high DPI.
* Changes in the NewVersionAvailableForm, for showing the labels correctly in high DPI, besides removing some event handlers no longer called by the form.
* Small improvement in the GitHubReleaseProvider class, to have one static readonly HttpClient instance created.
* Fixing the height of the header box in the Grouper control, besides replacing tabs with spaces.
* Addressing some warnings in the Common project related to the app.config mapping and binding redirects.

### DIFF
--- a/src/Common/Common.csproj
+++ b/src/Common/Common.csproj
@@ -3,7 +3,7 @@
     <ProjectGuid>{DE07DBEB-D772-49BA-BCEB-A7CE29308AE3}</ProjectGuid>
     <RootNamespace>ServiceBusExplorer.Common</RootNamespace>
     <AssemblyName>ServiceBusExplorer.Common</AssemblyName>
-    <TargetFramework>net462</TargetFramework>
+    <TargetFramework>net472</TargetFramework>
     <LangVersion>latest</LangVersion>
     <OutputPath>bin\$(Configuration)\</OutputPath>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
--- a/src/Common/app.config
+++ b/src/Common/app.config
@@ -28,7 +28,7 @@
       </dependentAssembly>
       <dependentAssembly>
         <assemblyIdentity name="System.Runtime.CompilerServices.Unsafe" publicKeyToken="b03f5f7f11d50a3a" culture="neutral" />
-        <bindingRedirect oldVersion="0.0.0.0-4.0.5.0" newVersion="4.0.5.0" />
+        <bindingRedirect oldVersion="0.0.0.0-4.0.5.0" newVersion="4.0.6.0" />
       </dependentAssembly>
       <dependentAssembly>
         <assemblyIdentity name="System.Buffers" publicKeyToken="cc7b13ffcd2ddd51" culture="neutral" />
--- a/src/Common/app.config
+++ b/src/Common/app.config
@@ -28,7 +28,7 @@
       </dependentAssembly>
       <dependentAssembly>
         <assemblyIdentity name="System.Runtime.CompilerServices.Unsafe" publicKeyToken="b03f5f7f11d50a3a" culture="neutral" />
-        <bindingRedirect oldVersion="0.0.0.0-4.0.5.0" newVersion="4.0.6.0" />
+        <bindingRedirect oldVersion="0.0.0.0-6.0.0.0" newVersion="6.0.0.0" />
       </dependentAssembly>
       <dependentAssembly>
         <assemblyIdentity name="System.Buffers" publicKeyToken="cc7b13ffcd2ddd51" culture="neutral" />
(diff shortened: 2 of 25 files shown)


ErikMogensen:
Nice. I only took a brief look, but it looks good to me.

I don't think upgrading to .NET Framework 4.7.2 is a problem. We'll release it in a new major version once this PR is merged.

sean-feldman-sh:
+1 for gradually releasing. Less chances of something going awfully bad.

+1 for .NET 4.7.2 bump. I wish we'd go with the last version, 4.8.x but older and still supported Server OS don't support it IIRC.

Thank you, @jmoralesv

jmoralesv:
@ErikMogensen @sean-feldman-sh thank you! In that case, I will move this pull request as ready for review, do we expect more approvals? How is the merge process?

On the other hand, .NET Framework 4.7.2. is still supported and [doesn't have an end date yet](https://learn.microsoft.com/en-us/lifecycle/products/microsoft-net-framework), so it is perfectly fine to use it.

For .NET Framework 4.8, it looks like for Windows Server 2019[ it has to be installed (4.7.2 is included by default)](https://learn.microsoft.com/en-us/dotnet/framework/install/on-server-2019); whereas [Windows Server 2022 already includes 4.8](https://learn.microsoft.com/en-us/dotnet/framework/install/on-server-2022).

SeanFeldman (approved):

ErikMogensen (approved):
Great and thanks for the cleanups!

ErikMogensen:
> +1 for .NET 4.7.2 bump. I wish we'd go with the last version, 4.8.x but older and still supported Server OS don't support it IIRC.

I see that Windows Server 2012 supports .NET 4.8. @SeanFeldman, is there a reason besides using the latest version that you wish we' use 4.8.x?

SeanFeldman:
.NET 4.8.x is the last version that Microsoft will release and maintain for the full framework. That's why.
But .NET 4.8.x is not supported on Windows Server 2016, which still has an extended warranty till 2027 and is likely to be used in the wild (I know of some clients that do).

jmoralesv:
Thanks @ErikMogensen @SeanFeldman for your approvals! Glad to see this PR merged 😄 

I'll publish a second one with more forms and code updated these days.

ErikMogensen:
.NET 4.8 is [supported on Windows Server 2016](https://learn.microsoft.com/en-us/dotnet/framework/install/on-windows-10#net-framework-48). However, that article mentions that it is not supported on some older Windows 10 versions so let's stick to 4.7.2 for now.

ErikMogensen:
Also, as mentioned, some Windows versions require an installation to get to .NET 4.8 which may be a problem.

sean-feldman-sh:
Windows 2012/R2 what I meant, not 2016.
Extended support till [October 13, 2026](https://techcommunity.microsoft.com/t5/windows-it-pro-blog/windows-server-2012-r2-extended-security-updates/ba-p/3976610)

ErikMogensen:
Not that it really matters but [.NET 4.8 is supported on Windows Server 2012](https://learn.microsoft.com/en-us/dotnet/framework/install/on-windows-8#net-framework-48) and on [Windows Server 2012 R2](https://learn.microsoft.com/en-us/dotnet/framework/install/on-windows-8-1#net-framework-48).

ErikMogensen:
Great work @jmoralesv! 

However when running the current main branch, which contains this PR I see that the tree view has gotten more compressed as shown in the figure below. The left one is running on an older commit. 
image

It may be my personal view, but I liked the old one better.  What do you think?

Also, the text is a bit blurry on the right side. This is on a screen with 100% Scale. On a screen using 150% Scale the text looks good. I am not sure if it is the Scale setting though. Do you know what causes this?

jmoralesv:
Hi @ErikMogensen that's correct, I can see a similar situation in Windows 10:
Windows10

I have at least one commit that can resolve this in my original commits, in a separate branch, which I can cherry pick from and add it to this PR: #800 

I will review my original commits and see if that commit and others can be added, but it will increase the number of changes in PR 800.

ErikMogensen:
> I have at least one commit that can resolve this in my original commits, in a separate branch, which I can cherry pick from and add it to this PR: #800
> 
> I will review my original commits and see if that commit and others can be added, but it will increase the number of changes in PR 800.

Excellent! Does this include the fixing the blurriness on some screens?

jmoralesv:
@ErikMogensen that's something I can't really see in the machine I have for testing, could you please share more details about the screen you're using to test with?

ErikMogensen:
It gets blurry, at least in my opinion, when switching to 100% scale on my laptop which has the following settings:
image

This is the current main branch:
image

This is the latest release:
image

ErikMogensen:
When changing the scaling to 150% on the same computer then it is the other way around.

PR #800, opened 2024-08-05
Second set of changes to enable high DPI in Service Bus Explorer

jmoralesv:
This pull request continues the work started in pull-request: #797 by using TableLayoutPanels across several forms and user controls, to improve the rendering and resizing of the inner controls in a high DPI monitor, so there is no need for custom paint and resize logic in some cases.

* Adding a TableLayoutPanel in the TimeSpanControl, to improve the scaling of the controls inside, so they can be displayed better. This also allows the control to have a bigger size in a container, and the controls can expand their size according to the container size. 
* Updating the HandleTopicControl to use the updated TimeSpanControl.
* Changes for the layout, sizes and position of the controls in the HandleTopicControl, HandleSubscriptionControl and HandleRuleControl. The last one has some temporary changes, because the resize of the controls is messy without a TableLayoutPanel.
* Adding a TableLayoutPanel to the HandleRuleControl, to improve the scaling and resizing of the child controls.
* Changes for showing the grouperCorrelationFilter control in the HandleRuleControl.
* Adding a TableLayoutPanel for the grouperCorrelationFilter control in the HandleRuleControl.
* Aligning the layout of the rest of Handle controls (HandleConsumerGroup, HandleEventHub, HandleNotificationHub, HandleQueue, HandleRelay, HandleSubscription, HandleTopic), so the groupers and others are in similar positions.
* Adding calls to SuspendDrawing, SuspendLayout, ResumeDrawing and ResumeLayout when rendering the groupers in the HandleRuleControl.
* Changes for rendering the headers of the list views in the Handle controls, including HandlePartition, ListenerControl, PartitionListenerControl, TestEventHub, TestQueue, TestRelay, TestSubscription, and TestTopic, by calling e.DrawText.
* Changes in the HandleNotificationHubControl, to use a table layout panel to improve the layout when rendering the form in a high DPI monitor. Because of this, the HandleNotificationHubControl_Resize method is no longer needed.
* Removing the HandleRuleControl_Resize method in the HandleRuleControl.


src/ServiceBusExplorer/Controls/HandleConsumerGroupControl.cs
```diff
@@ -501,16 +495,15 @@ private void propertyListView_DrawColumnHeader(object sender, DrawListViewColumn
...
-            e.Graphics.DrawLine(new Pen(SystemColors.ControlLightLight), startX, -1, endX, -1);
+            e.Graphics.DrawLine(SystemPens.ControlLightLight, startX, -1, endX, -1);
             // Bottom horizontal line
-            e.Graphics.DrawLine(new Pen(SystemColors.ControlDark), startX, e.Bounds.Height - 1, endX, e.Bounds.Height - 1);
+            e.Graphics.DrawLine(SystemPens.ControlDark, startX, e.Bounds.Height - 1, endX, e.Bounds.Height - 1);
             // Right vertical line
-            e.Graphics.DrawLine(new Pen(SystemColors.ControlDark), endX, -1, endX, e.Bounds.Height + 1);
-            var roundedFontSize = (float)Math.Round(e.Font.SizeInPoints);
-            var bounds = new RectangleF(e.Bounds.X + 4, (e.Bounds.Height - 8 - roundedFontSize) / 2, e.Bounds.Width, roundedFontSize + 6);
-            e.Graphics.DrawString(e.Header.Text, e.Font, new SolidBrush(SystemColors.ControlText), bounds);
```
ErikMogensen, line 513: It looks like the` e.Header.Text` isn't shown. Is it showed in a different way? Or isn't it needed?

jmoralesv: @ErikMogensen if I'm not wrong, e.DrawText() accomplish the same thing, which is the same change in the rest of the Handle controls.

I think this would render the column headers in the list view, I tested with event hubs, queues, subscriptions and topics and it works fine. I don't know how to test it with this ConsumerGroup, could you please tell me how I can test this code with a ConsumerGroup?

ErikMogensen:
I want to wait with this PR until we have a solution for the blurriness for text in some scenarios as mentioned in #797.

PR #801, opened 2024-08-08
Replace Nunit with xUnit

asos-gurpreetsingh:


SeanFeldman (changes requested):
Thank you, @asos-gurpreetsingh. The unit tests in this project are... insufficient to say the least.
Looking at the changes to NUnit, I'm inclined to decline this. Here's why

This new NUnit API is attempting to mimic [Fluent Assertions](https://fluentassertions.com) but looks off. Take the following example:

```
Assert.That(false, Is.EqualTo(nonExistingValueAsFalse));
``

Every assertion has to have `Assert.That()` as a container method. With Fluent Assertion, the same would be more readable.

```
nonExistingValueAsFalse.Should().BeFalse();
```

The benefits IMO are
1. More fluid
2. No "container method"
3. Reads better for humans

I can support Fluent Assertions if you'd like. However, replacing an existing clunky NUnit API with another variant is not worth it.

**Comment by asos-gurpreetsingh** [comment]
Thanks @SeanFeldman , are we saying that if we update the Nunit package then we want to use fluent assertions otherwise we will never update Nunit package ?

**Comment by SeanFeldman** [comment]
Is the new API mandatory for the latest version of NUnit?

**Comment by asos-gurpreetsingh** [comment]
The old methods were not available when I updated

**Comment by SeanFeldman** [comment]
Thanks, @asos-gurpreetsingh.
If you're already at it, we might as well swap NUnit with XUnit. It's a more commonly used library this days, including MSFT projects.

File: .github/workflows/build-test.yml [hunk 1]
```diff
@@ -6,42 +6,33 @@ on:
...
   SOLUTION_FILE_PATH: .\src
   BUILD_CONFIGURATION: Release
 
 jobs:
   build_test_publish:
     name: Build and run unit tests
-    runs-on: windows-2022
-    env:
-      NUnitVersion: '3.16.3'
+    runs-on: windows-latest
```
ErikMogensen, line 26: I think it is better to fix the OS version. That way you are in control of it. That trumps the disadvantage of having to update it every third year or so.

.github/workflows/build-test.yml
```diff
@@ -50,37 +41,18 @@ jobs:
...
-
-        $notificationHubsInfoFile = "$env:GITHUB_WORKSPACE\src\NotificationHubs\Properties\AssemblyInfo.cs"
-        .\SetVersion -FileName $notificationHubsInfoFile -PropertyName 'AssemblyFileVersion' -Version ${{env.NUMBER_VERSION}}      
-        .\SetVersion -FileName $notificationHubsInfoFile -PropertyName 'AssemblyInformationalVersion' -VersionString ${{env.ReleaseVersion}}   
-
     - name: Build
-      run: |
-        msbuild /m /property:Configuration=${{env.BUILD_CONFIGURATION}},FileVersion=${{env.NUMBER_VERSION}},InformationalVersion=${{inputs.release-version}} ${{env.SOLUTION_FILE_PATH}}
+      working-directory: ${{env.SOLUTION_FILE_PATH}}
+      run: dotnet build -c ${{env.BUILD_CONFIGURATION}} /p:AssemblyVersion=${{env.NUMBER_VERSION}} /p:InformationalVersion=${{inputs.release-version}}
```
ErikMogensen, line 51: I doubt that dotnet build sets File version and Informational version on .NET Framework assemblies. Have you checked this?

.github/workflows/build-test.yml
sean-feldman-sh: Is there anything specific from Windows 2022 OS we require? I'm actually in favor of running with the latest. If it doesn't cause anything to fail during the build, were covered. If the latest will cause something to break, we'll investigate and fix it. So +1 to go with the latest for the build pipeline.

src/ServiceBusExplorer/Controls/TrackBar.cs
```diff
@@ -627,9 +627,6 @@ public int Minimum
 
 				if (minimum > maximum)
 					maximum = minimum;
-				if (minimum > value)
```
SeanFeldman, line 630: This is not part of xUnit migration. Let's make it a separate PR.

src/ServiceBusExplorer/ServiceBusExplorer.csproj
```diff
@@ -24,6 +24,7 @@
     <IsWebBootstrapper>false</IsWebBootstrapper>
     <UseApplicationTrust>false</UseApplicationTrust>
     <BootstrapperEnabled>true</BootstrapperEnabled>
+	<GenerateResourceUsePreserializedResources>true</GenerateResourceUsePreserializedResources>
```
ErikMogensen, line 27: Why was this line added?

src/ServiceBusExplorer/Controls/TrackBar.cs
```diff
@@ -651,9 +648,6 @@ public int Maximum
 			{
 				maximum = value;
 
-				if (maximum < value)
```
SeanFeldman, line 654: This is not part of xUnit migration. Let's make it a separate PR.

src/ServiceBusExplorer/ServiceBusExplorer.csproj
```diff
@@ -549,6 +550,7 @@
     <PackageReference Include="Microsoft.Azure.Services.AppAuthentication" Version="1.6.2" />
     <PackageReference Include="Microsoft.CSharp" Version="4.7.0" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
+    <PackageReference Include="System.Resources.Extensions" Version="8.0.0" />
```
ErikMogensen, line 553: I also wonder what this line does?

src/ServiceBusExplorer/UIHelpers/CustomPropertyDescriptor.cs
```diff
@@ -236,21 +236,9 @@ public override bool IsLocalizable
...
-            {
-                var attr = (DefaultValueAttribute)attributes.FirstOrDefault(a => a is DefaultValueAttribute);
-                if (attr == null)
-                {
-                    attributes.RemoveAll(a => a is DefaultValueAttribute);
-                }
-                // ReSharper disable once RedundantAssignment
-                attr = new DefaultValueAttribute(value); // TODO: what's this doing?
-            }
+            get => attributes.OfType<DefaultValueAttribute>().FirstOrDefault()?.Value;
```
SeanFeldman, line 239: This is not part of xUnit migration. Let's make it a separate PR.

ErikMogensen (commented):
I like the switch to xUnit and the new assertions.

SeanFeldman (changes requested):
A couple of minor things. The xUnit changes look good.

src/ServiceBusExplorer/ServiceBusExplorer.csproj
SeanFeldman: > Provides classes which read and write resources in a format that supports non-primitive objects.

I don't see it used anywhere though.

SeanFeldman: If this is not part of xUnit migration then let's make it a separate PR.

.github/workflows/build-test.yml
ErikMogensen: As far as I know there isn't anything in the build that requires it to run on Windows 2022, but investigating an issue which is due to running on another OS version gets more complicated since you'll start by checking the updated source code.

.github/workflows/build-test.yml
asos-gurpreetsingh: I can see the assembly version set when I build it on my machine with same command

.github/workflows/build-test.yml
asos-gurpreetsingh: reverted to windows-2022

src/ServiceBusExplorer/ServiceBusExplorer.csproj
asos-gurpreetsingh: dotnet build command fails while complaining that it cannot find this namespace at run time
image

src/ServiceBusExplorer/UIHelpers/CustomPropertyDescriptor.cs
asos-gurpreetsingh: reverted

SeanFeldman (approved):

SeanFeldman:
I don't recall wherever PR build should have a version or not. @ErikMogensen do you remember?

Screenshot_2024-08-12-07-29-57-87_40deb401b9ffe8e1df2f1cc5ba480b12

asos-gurpreetsingh:
> I don't recall wherever PR build should have a version or not. @ErikMogensen do you remember?
> 
> Screenshot_2024-08-12-07-29-57-87_40deb401b9ffe8e1df2f1cc5ba480b12

I added default input for workflow manual build that is what shows version number here, without this i cannot test build in feature branch, it could not find version release-version which is set as env variable.

ErikMogensen:
> I don't recall wherever PR build should have a version or not. @ErikMogensen do you remember?

They do not have a version number so this is correct.

.github/workflows/build-test.yml
ErikMogensen: When I run: `dotnet build -c RELEASE /p:AssemblyVersion=0.1.2.3 /p:InformationalVersion=0.0.1.1` in the src directory using this branch the .NET Standard assemblies get the same Assembly version as the File version. The Assembly versions should be 1.0.0.0 on the assemblies and the File Versions should be the release version. 

The built .NET Standard assemblies lack File version and Product version. This can be checked by opening them using dotPeek. This is ServiceBusExplorer.Common:
image

Below is how the assembly in the latest release looks like in dotPeek :
image

asos-gurpreetsingh: Actually i am not able to get the versions set by dotnet build or msbuild. I will try to figure it out and update the build task

asos-gurpreetsingh: I am 100% sure it worked last time is tested, i hope i did not  build something else :)

src/ServiceBusExplorer/ServiceBusExplorer.csproj
asos-gurpreetsingh: I got it working using .props file for solution.
image

asos-gurpreetsingh: these are just test values i used for version, actual version will be supplied by the build task

.github/workflows/build-test.yml
asos-gurpreetsingh: It seems to have worked only for .netstandard projects and that was the assembly i checked. I have it working for all now.

asos-gurpreetsingh:
So what is next ? who will complete the PR ?

SeanFeldman:
> So what is next ? who will complete the PR ?

Patience. This is not paid work, and people have their own priorities :)

src/ServiceBusExplorer.Tests/ServiceBusExplorer.Tests.csproj
```diff
@@ -27,8 +27,15 @@
...
   <ItemGroup>
-    <PackageReference Include="NUnit" Version="3.14.0" />
-    <PackageReference Include="NUnit3TestAdapter" Version="4.5.0" />
+    <PackageReference Include="FluentAssertions" Version="6.12.0" />
     <PackageReference Include="WindowsAzure.ServiceBus" Version="7.0.1" />
+    <PackageReference Include="xunit" Version="2.9.0" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.8.2">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
```
SeanFeldman, line 36: nit: KISS
```suggestion
    <PackageReference Include="xunit.runner.visualstudio" Version="2.8.2" PrivateAssets="all" IncludeAssets="runtime; build; native; contentfiles; analyzers; buildtransitive" />
    </PackageReference>
```

src/ServiceBusExplorer.Tests/ServiceBusExplorer.Tests.csproj
```diff
@@ -27,8 +27,15 @@
...
+  </ItemGroup>
+  <ItemGroup>
+    <Folder Include="Properties\" />
   </ItemGroup>
```
SeanFeldman, line 40: This can be removed, right?

```suggestion
```

.github/workflows/build-test.yml
```diff
@@ -50,37 +42,18 @@ jobs:
...
-
-        $notificationHubsInfoFile = "$env:GITHUB_WORKSPACE\src\NotificationHubs\Properties\AssemblyInfo.cs"
-        .\SetVersion -FileName $notificationHubsInfoFile -PropertyName 'AssemblyFileVersion' -Version ${{env.NUMBER_VERSION}}      
-        .\SetVersion -FileName $notificationHubsInfoFile -PropertyName 'AssemblyInformationalVersion' -VersionString ${{env.ReleaseVersion}}   
-
     - name: Build
-      run: |
-        msbuild /m /property:Configuration=${{env.BUILD_CONFIGURATION}},FileVersion=${{env.NUMBER_VERSION}},InformationalVersion=${{inputs.release-version}} ${{env.SOLUTION_FILE_PATH}}
+      working-directory: ${{env.SOLUTION_FILE_PATH}}
+      run: dotnet build -c ${{env.BUILD_CONFIGURATION}} -p:GenerateAssemblyInfo=true,AssemblyVersion=${{env.NUMBER_VERSION}},InformationalVersion=${{inputs.release-version}}
```
SeanFeldman, line 52: This begs a unit test to validate we set the release version only if you want to add that test.

SeanFeldman (approved):

SeanFeldman:
@asos-gurpreetsingh, approved with a few suggestions. We usually try to have two reviewers on something that's a minor change. With the build version change on top of xUnit, I'll let @ErikMogensen get another pass. The good news is that this is looking good and close to completion. Thank you.

asos-gurpreetsingh:
> > So what is next ? who will complete the PR ?
> 
> Patience. This is not paid work, and people have their own priorities :)

Apologies, I did not mean to pressure, I just wanted to know what's next.

.github/workflows/build-test.yml
asos-gurpreetsingh: Good idea actually. I will put something in

asos-gurpreetsingh: Were you thinking about a unit test ? how would that go ?
I was thinking a unit test that check the version should be set to expected version that comes from pipeline/environment variable, but then it would fail locally because we don't have it set there.

ErikMogensen:
I tested to build this branch using `dotnet build -c RELEASE -p:GenerateAssemblyInfo=true,AssemblyVersion=0.1.2.3,InformationalVersion=0.0.1.1`. It sets File version to the same value as the Assembly version, which we don't want. Assembly version should be set per assembly. File version should be the same on all assemblies created by the build.

SeanFeldman:
@asos-gurpreetsingh, do you plan to complete this one? Erik pointed out one [last thing](https://github.com/paolosalvatori/ServiceBusExplorer/pull/801#issuecomment-2294802536) that needs attention.

asos-gurpreetsingh:
> @asos-gurpreetsingh, do you plan to complete this one? Erik pointed out one [last thing](https://github.com/paolosalvatori/ServiceBusExplorer/pull/801#issuecomment-2294802536) that needs attention.

Hi Yes I will pick this weekend, i was out on holiday

asos-gurpreetsingh:
> I tested to build this branch using `dotnet build -c RELEASE -p:GenerateAssemblyInfo=true,AssemblyVersion=0.1.2.3,InformationalVersion=0.0.1.1`. It sets File version to the same value as the Assembly version, which we don't want. Assembly version should be set per assembly. File version should be the same on all assemblies created by the build.

In your test you have used very different version numbers, if our objective is to use something like 1.2.3.0 as assembly version and 1.2.3.566 as file version then this command would do that. But if we really want to set different versions like in you example then we can use another Property "AssemblyFileVersion" for that

ErikMogensen:
Yes, we want to be able to set different versions. Thinking of this, I want the assembly version to be set using a file belonging to the project of the affected assembly instead of being set in the pipeline.

ErikMogensen:
There is also an issue with the build.

asos-gurpreetsingh:
> Yes, we want to be able to set different versions. Thinking of this, I want the assembly version to be set using a file belonging to the project of the affected assembly instead of being set in the pipeline.

Can I ask why ?

.github/workflows/build-test.yml
asos-gurpreetsingh: I have now added two unit tests to check the file version and assembly version

src/ServiceBusExplorer.Tests/AssemblyVersionTests.cs
```diff
@@ -0,0 +1,42 @@
...
+using Xunit;
+
+namespace ServiceBusExplorer.Tests
+{
+    public class AssemblyVersionTests
+    {
+        private string expectedVersion;
+        private string expectedFileVersion;
+
+        public AssemblyVersionTests()
```
SeanFeldman, line 14: This is a great automated gate!

nit: It's not just an assembly version verification but an assembly and file version. Class/file name could reflect that.

asos-gurpreetsingh: now split into two files with suitable names

sean-feldman-sh: Also works 🙂

ErikMogensen:
> > Yes, we want to be able to set different versions. Thinking of this, I want the assembly version to be set using a file belonging to the project of the affected assembly instead of being set in the pipeline.
> 
> Can I ask why ? We could use git-version to drive version numbers.

Of course. Assembly version is used by the .NET runtime to determine and verify it is loading the intended assembly. The assembly number of a dependent assembly is stored in the manifest of the calling assembly. By default it will only load that assembly.

File version may be used for installers to determine which version an assembly has so the numbers have different purposes. 

By setting all the assemblies built in a build to have the same file version it is easy to tell for an installer and a human which build a certain assembly belongs to. Typically you do not want to change the assembly version in every build since that would make assemblies between two different builds incompatible. The exception to this if you have made changes in the assembly that lost its backward combability, then it should increase its assembly number.

Does this PR still set the assembly version to the same value on all assemblies?

asos-gurpreetsingh:
> > > Yes, we want to be able to set different versions. Thinking of this, I want the assembly version to be set using a file belonging to the project of the affected assembly instead of being set in the pipeline.
> > 
> > 
> > Can I ask why ? We could use git-version to drive version numbers.
> 
> Of course. Assembly version is used by the .NET runtime to determine and verify it is loading the intended assembly. The assembly number of a dependent assembly is stored in the manifest of the calling assembly. By default it will only load that assembly.
> 
> File version may be used for installers to determine which version an assembly has so the numbers have different purposes.
> 
> By setting all the assemblies built in a build to have the same file version it is easy to tell for an installer and a human which build a certain assembly belongs to. Typically you do not want to change the assembly version in every build since that would make assemblies between two different builds incompatible. The exception to this if you have made changes in the assembly that lost its backward combability, then it should increase its assembly number.
> 
> Does this PR still set the assembly version to the same value on all assemblies?

Thanks, @ErikMogensen . I understand now why one could want different versions but I am not sure we SBE uses it with that intent. But with current state of build task "dotnet build ....." allows us to use different versions for file and assembly but that will be applied to all assemblies. I don't think this we ever set up to use different file versions for each project. That can be done by setting those in the project files. Secondly if that is what we want then we can pick this up in another change and not this one.

ErikMogensen:
> Thanks, @ErikMogensen . I understand now why one could want different versions but I am not sure we SBE uses it with that intent. 
We should probably have updated the assembly numbers in some PRs.

> But with current state of build task "dotnet build ....." allows us to use different versions for file and assembly but that will be applied to all assemblies. I don't think this we ever set up to use different file versions for each project. That can be done by setting those in the project files. 
We don't want to have different file versions for assemblies in the same build so that's fine. 

> Secondly if that is what we want then we can pick this up in another change and not this one.
All work on this tool is voluntarily so there is quite a big risk that such a change will never be created. 

What is the problem with using msbuild for building?

asos-gurpreetsingh:
> > Thanks, @ErikMogensen . I understand now why one could want different versions but I am not sure we SBE uses it with that intent.
> 
> We should probably have updated the assembly numbers in some PRs.
> 
> > But with current state of build task "dotnet build ....." allows us to use different versions for file and assembly but that will be applied to all assemblies. I don't think this we ever set up to use different file versions for each project. That can be done by setting those in the project files.
> 
> We don't want to have different file versions for assemblies in the same build so that's fine.
> 
> > Secondly if that is what we want then we can pick this up in another change and not this one.
> 
> All work on this tool is voluntarily so there is quite a big risk that such a change will never be created.
> 
> What is the problem with using msbuild for building?

Apologies if I am missing something. when you say "I want the assembly version to be set using a file belonging to the project of the affected assembly instead of being set in the pipeline. " what is that to do with MSBuild or dotnet build ? and secondly "using a file belonging to the project of the affected assembly" sounds contradicting to "We don't want to have different file versions for assemblies". Could you please elaborate. Thanks

ErikMogensen:
> Apologies @ErikMogensen if I am missing something. when you say "I want the assembly version to be set using a file belonging to the project of the affected assembly instead of being set in the pipeline. " what is that to do with MSBuild or dotnet build ?

I assumed dotnet build cannot handle assembly version set in a project since you changed the way assembly versions were set. But if it is possible to set assembly versions using files belonging to each project and still use dotnet build, it would be great.

> secondly "using a file belonging to the project of the affected assembly" sounds contradicting to "We don't want to have different file versions for assemblies". Could you please elaborate. Thanks

It is about File version vs Assembly version. We want to have the same File version on all assemblies created by the build. We want Assembly version set separately for each assembly.

Just ask again if it is unclear 😄

asos-gurpreetsingh:
> > Apologies @ErikMogensen if I am missing something. when you say "I want the assembly version to be set using a file belonging to the project of the affected assembly instead of being set in the pipeline. " what is that to do with MSBuild or dotnet build ?
> 
> I assumed dotnet build cannot handle assembly version set in a project since you changed the way assembly versions were set. But if it is possible to set assembly versions using files belonging to each project and still use dotnet build, it would be great.
> 
> > secondly "using a file belonging to the project of the affected assembly" sounds contradicting to "We don't want to have different file versions for assemblies". Could you please elaborate. Thanks
> 
> It is about File version vs Assembly version. We want to have the same File version on all assemblies created by the build. We want Assembly version set separately for each assembly.
> 
> Just ask again if it is unclear.

Thanks @ErikMogensen . In that case file version can be set by dotnet build, but setting different assembly versions for each assembly would mean updating project files manually when needed. What assembly version should we set to start with ? when would it be changed (on what basis)

ErikMogensen:
> Thanks @ErikMogensen . In that case file version can be set by dotnet build, but setting different assembly versions for each assembly would mean updating project files manually when needed. 

Great!

> What assembly version should we set to start with ? 

Look with the ones that are in the main branch. (I believe all are 1.0.0,0 which is most likely not right.) Since this PR changed the .NET version, increment the major version digit (the first one). 

> when would it be changed (on what basis)

They should be increased when there is new release of the assembly with an interface that is not backwards compatible. By interface I mean, public methods, properties, interfaces, events and delegates. (I may have forgotten one or two.) Also, if the runtime requirements change I consider that a change that should increase the version number.

asos-gurpreetsingh:
@ErikMogensen I have set set the assembly versions to 1.0.0.1 in all projects and build script will provide the file version. I have updated the tests to reflect that too. Please check if this works as you expected. Could you also confirm that the publish process is not affected by this. (i have checked and it does not look like it should be affected, but please check)

ErikMogensen:
Well done @asos-gurpreetsingh !

PR #805, opened 2024-11-04
Show error message when a fatal error occurs

ErikMogensen:
This happens if the user configuration contains invalid XML, for instance.

### DIFF
--- a/src/ServiceBusExplorer/Program.cs
+++ b/src/ServiceBusExplorer/Program.cs
@@ -70,7 +70,7 @@ namespace ServiceBusExplorer
             }
             catch (Exception e)
             {
-                MessageBox.Show(text: $"A fatal exception occurred. ServiceBusExplorer will now shutdown. \n\n{e.Message}.", 
+                MessageBox.Show(text: $"A fatal exception occurred. ServiceBusExplorer will now shut down. \n\n{e.Message}.", 
                     caption: "ServiceBusExplorer", 
                     buttons: MessageBoxButtons.OK,
                     icon: MessageBoxIcon.Stop);
--- a/src/ServiceBusExplorer/Program.cs
+++ b/src/ServiceBusExplorer/Program.cs
@@ -68,10 +68,12 @@ namespace ServiceBusExplorer
                     Application.Run(new MainForm(helpText));
                 }
             }
-            // ReSharper disable EmptyGeneralCatchClause
-            catch (Exception)
-            // ReSharper restore EmptyGeneralCatchClause
+            catch (Exception e)
             {
+                MessageBox.Show(text: $"A fatal exception occurred. ServiceBusExplorer will now shutdown. \n\n{e.Message}.", 
+                    caption: "ServiceBusExplorer", 
+                    buttons: MessageBoxButtons.OK,
+                    icon: MessageBoxIcon.Stop);
             }
         }
         


src/ServiceBusExplorer/Program.cs
```diff
@@ -68,10 +68,12 @@ static void Main(string[] args)
                     Application.Run(new MainForm(helpText));
                 }
             }
-            // ReSharper disable EmptyGeneralCatchClause
-            catch (Exception)
-            // ReSharper restore EmptyGeneralCatchClause
+            catch (Exception e)
             {
+                MessageBox.Show(text: $"A fatal exception occurred. ServiceBusExplorer will now shutdown. \n\n{e.Message}.",
```
SeanFeldman, line 73:
```suggestion
                MessageBox.Show(text: $"A fatal exception occurred. ServiceBusExplorer will now shut down. \n\n{e.Message}.", 
```

SeanFeldman: Is it possible to copy the stack trace to the clipboard? It's be nice but not a showstopper.

SeanFeldman (commented):
A couple of comments.
Didn't know SBE fails upon startup. Would it be because of invalid settings? If that the case, perhaps we should be more explicit.

ErikMogensen:
> Didn't know SBE fails upon startup. Would it be because of invalid settings? If that the case, perhaps we should be more explicit.

If I add a dash to the configuration tag in the user config file, SBE just silently dies. This PR makes it say:
image

Originally I showed the stack trace but it was huge and would scare many users.

ErikMogensen:
The stack could automatically be copied to the clipboard, but that would be unexpected and sometimes not desired. Adding a button for optionally copying the stack, would as far as I know require a custom dialog box, which is more than I had in mind for this PR. But anyone is free to add  that functionality in the future... 😃

SeanFeldman:
> The stack could automatically be copied to the clipboard, but that would be unexpected and sometimes not desired. 

Why would it be undesired? Also, if you say in the error that the stacktrace has been copied to the clipboard, it's no longer unexpected 🙂 

Personally, I really dislike taking screenshots of errors because there's no other way to re-read that error message. And sure thing I will not remember which line and column there was an issue with.

SeanFeldman (approved):

ErikMogensen:
> Why would it be undesired? Also, if you say in the error that the stacktrace has been copied to the clipboard, it's no longer unexpected 🙂

What if I have a Entra client secret in the clipboard, start SBE and then I would lose the secret and have a stack trace there instead. 

> Personally, I really dislike taking screenshots of errors because there's no other way to re-read that error message. And sure thing I will not remember which line and column there was an issue with.

I agree. I would like to see it writing the stack to a log file. In the error message it should show the path to the file.

SeanFeldman:
> What if I have a Entra client secret in the clipboard, start SBE and then I would lose the secret and have a stack trace there instead.

Clipboard on Windows 10 and 11 can retain more than one entry.

ErikMogensen:
> Clipboard on Windows 10 and 11 can retain more than one entry.

Yes, but I believe there is a lot of people who don't use it. Also, I have never used an application that replaced the contents of the clipboard on its own accord.

SeanFeldman:
Then, you'll be forced to memorize the line and the position. Fun.

Issue #804, opened 2024-10-05
The Repair and Resubmit form does not look good


ErikMogensen:
The Repair and Resubmit form does not look good on 100% and a screen with 3440 x 1440 resolution. This is on Windows 11. 

![image](https://github.com/user-attachments/assets/de0d0505-1e49-4e27-8f4d-1e4e6be422f0)

This is a regression caused by the 4K adaption. As mentioned in another issue/PR there are other issues with that PR too.

SeanFeldman:
4K or High DPI (#797) PR?

ErikMogensen:
My bad. It's #797.

ErikMogensen:
The Submit form is not good either. 
![image](https://github.com/user-attachments/assets/e55ed287-66a7-4fa7-aa8e-eb14d4f64e16)

PR #806, opened 2024-11-04
Revert high dpi changes

ErikMogensen:
As mentioned in #804, there are some GUI things that don't look good after the changes for high DPI. Therefore this change reverts those changes.

Sorry @jmoralesv. I hope you will find the time to get this working for those scenarios.


SeanFeldman:
@ErikMogensen, could you please squash those two commits? A revert is better to be a single commit. Thanks.

ErikMogensen:
> @ErikMogensen, could you please squash those two commits? A revert is better to be a single commit. Thanks.

Don't you think it is enough to do it when merging it into the main branch? The end result will be same, as far as I can tell.

SeanFeldman:
When looking at the PRs, merged or not, they retain all the commits. When a PR is "revert", it's better to have a single commit.

ErikMogensen:
> When looking at the PRs, merged or not, they retain all the commits. When a PR is "revert", it's better to have a single commit.

That's true. Looks like I have to abort this PR and create a new one.

PR #807, opened 2024-11-04
Revert the High DPI changes

ErikMogensen:
As mentioned in https://github.com/paolosalvatori/ServiceBusExplorer/issues/804, there are some GUI things that don't look good after the changes for high DPI. Therefore this change reverts those changes.

Sorry @jmoralesv. I hope you will find the time to get this working for those scenarios.

Second try as the first PR had two commits.


src/Common/Common.csproj
```diff
@@ -3,7 +3,7 @@
     <ProjectGuid>{DE07DBEB-D772-49BA-BCEB-A7CE29308AE3}</ProjectGuid>
     <RootNamespace>ServiceBusExplorer.Common</RootNamespace>
     <AssemblyName>ServiceBusExplorer.Common</AssemblyName>
-    <TargetFramework>net472</TargetFramework>
+    <TargetFramework>net462</TargetFramework>
```
SeanFeldman, line 6: Are there any concerns about reverting the target framework? In theory, this change shouldn't impact UI at all.

ErikMogensen: There were some changes in the original PR that I would have liked to keep, but to keep it clean and as I am pressed for time I didn't change them. However, I think we should stick to NET 4.7.2 for all NET Framework assemblies, since the XUnit PR may have assemblies that are dependent upon that version. 

Hopefully there is a new High DPI PR and that will require 4.7.2 anyways.

How do we handle this?

SeanFeldman: I'd say revert everything but the TF. Keep it as `net472`.

ErikMogensen: @SeanFeldman, do you want these changes to be merged as one? Since that, as far as I know, requires me to create a a new branch new PR and , I'd prefer to keep it as two commits. Otherwise we'll have three PRs for the same issue.

sean-feldman-sh: Proceed with what will be simpler for you.

ErikMogensen: Nice. We should be good to go then.

SeanFeldman (approved):

PR #808, opened 2024-11-07
Trying to fix Issue #775

### LINKED ISSUE #775: Export/Import of topic
When exporting a topic that has a filter on SessionId all other filter properties are exported as empty.
When trying to import this results in an error because a filter with an empty CorrelationId is not allowed.
Deleting CorrelationId from the exported XML file, results in an error when trying to load the exported XML file because the CorrelationId is "missing".

NaorisKong:
fix #775 
Trying to fix the Topic import bug.
Added const string of correlation filter name.
Added CorrelationFilter initialization logic in CreateFilter function.

### DIFF
--- a/src/Common/Helpers/ImportExportHelper.cs
+++ b/src/Common/Helpers/ImportExportHelper.cs
@@ -21,6 +21,11 @@
 
 #region References
 
+using Microsoft.Azure.NotificationHubs;
+using Microsoft.ServiceBus;
+using Microsoft.ServiceBus.Management;
+using Microsoft.ServiceBus.Messaging;
+using ServiceBusExplorer.Utilities.Helpers;
 using System;
 using System.Collections.Generic;
 using System.ComponentModel;
@@ -33,10 +38,6 @@ using System.Text;
 using System.Threading.Tasks;
 using System.Xml;
 using System.Xml.Linq;
-using Microsoft.Azure.NotificationHubs;
-using ServiceBusExplorer.Utilities.Helpers;
-using Microsoft.ServiceBus;
-using Microsoft.ServiceBus.Messaging;
 
 #endregion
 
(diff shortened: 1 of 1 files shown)


src/Common/Helpers/ImportExportHelper.cs
```diff
@@ -1348,7 +1355,17 @@ private Filter CreateFilter(XElement filter)
             }
             if (filter.Name == string.Format(NodeNameFormat, Namespace, CorrelationFilterEntity))
             {
-                ruleFilter = new CorrelationFilter(propertyValue[CorrelationId] as string);
+                ruleFilter = new CorrelationFilter()
+                {
+                    ContentType = !string.IsNullOrWhiteSpace(propertyValue[ContentType] as string) ? propertyValue[ContentType] as string : null,
```
SeanFeldman, line 1360: Question: is it necessary to use ternary condition if it's either a value or a null? Could this (one example) be it?

```
ContentType = propertyValue[ContentType] as string,
```

NaorisKong: If use the code 
`ContentType = propertyValue[ContentType] as string`
May cause the exception "Exception: The argument  contentType is null or white space."
Because when the xml be like this "<ContentType></ContentType>", use propertyValue[ContentType] will get a empty string and I guess the construction function of CorrelationFilter can`t accept a empty string.

SeanFeldman: Indeed. This is to avoid empty string. Thank you for elaboration. In that case, could you make it a local function and pass the property name? On the local function you could put XML documentation or a comment elaborating the rationale. Thank you.

NaorisKong: Ok, I will try to create a function to deal with it and add some comments.

NaorisKong: Added a extension method on Dictionary<string, object>.Think it`s cleaner.

src/Common/Helpers/ExtensionMethods.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    public static class ExtensionMethods
+    {
+        /// <summary>
+        /// this exthension method is used to get the string value from the dictionary
+        /// if the value is null or empty then it will return null
+        /// </summary>
+        /// <param name="propertyValue"></param>
+        /// <param name="propertyName"></param>
+        /// <returns></returns>
```
SeanFeldman, line 17:
```suggestion
        /// <summary>
        /// Get the string value from the dictionary. Return null when the value is null or empty.
        /// </summary>
```

PR #813, opened 2024-11-24
fixes an issue when connection string ending with ";" causes an exception

dileepsam:
If a connection string ends with a ";" semi colon the split function returns an empty string as the last entry causing an exception.

By removing those empty entries, exception can be avoided, and application works as expected.


SeanFeldman (commented):
Good change.
Thank you, @dileepsam.

PR #815, opened 2024-12-02
Implement issue 802

### LINKED ISSUE #802: Keep the subscriptions nodes showing and expanded when a topic is refreshed
I have the subscription nodes expanded and showing for a topic. When I refresh a topic, the subscription nodes collapse and I have to expand them to notice the refresh. The nodes show stay expanded.

NaorisKong:
Fixes #802 

just implement the feature when refresh the topic node.
not sure is it necessary to implement at other mode(Queues, Eventhhubs), as their Treeview structure was different.
not keep subscription node and rule node expaned, as it will add some ugly compare code in Foreach cycle and in daily use I just want keep subscriptions node expaned.

### DIFF
--- a/src/ServiceBusExplorer/Forms/MainForm.cs
+++ b/src/ServiceBusExplorer/Forms/MainForm.cs
@@ -1831,7 +1831,7 @@ namespace ServiceBusExplorer.Forms
             }
 
             //record the original subscriptions node is expanded
-            var originalSubscriptionsNode = selectedNode.Nodes.Find(SubscriptionEntities, false).First();
+            var originalSubscriptionsNode = selectedNode.Nodes.Find(SubscriptionEntities, false).FirstOrDefault();
             if (originalSubscriptionsNode != null)
             {
                 wasSubscriptionsNodeExpaned = originalSubscriptionsNode.IsExpanded;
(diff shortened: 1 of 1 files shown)


src/ServiceBusExplorer/Forms/MainForm.cs
```diff
@@ -1829,6 +1830,13 @@ private void RefreshIndividualTopic(TreeNode selectedNode)
                 return;
             }
 
+            //record the original subscriptions node is expanded
+            var originalSubscriptionsNode = selectedNode.Nodes.Find(SubscriptionEntities, false).First();
```
SeanFeldman, line 1834: If there's nothing, `.First()` will throw.

```suggestion
            var originalSubscriptionsNode = selectedNode.Nodes.Find(SubscriptionEntities, false).FirstOrDefault();
```

NaorisKong: yes, of course, my fault

src/ServiceBusExplorer/Forms/MainForm.cs
```diff
@@ -1817,6 +1817,7 @@ private void RefreshIndividualSubscription(SubscriptionDescription subscriptionD
         private void RefreshIndividualTopic(TreeNode selectedNode)
         {
             var wasTopicNodeExpanded = selectedNode.IsExpanded;
+            var wasSubscriptionsNodeExpaned = false;
```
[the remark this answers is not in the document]
NaorisKong: fixed

ErikMogensen (approved):
Very good!

Issue #686, opened 2022-10-02
Refactor the class ServiceBusHelper into multiple classes


ErikMogensen:
To enable migrating the code to use the latest SDK tracks for Relay, Notification hubs, Event hubs and Service bus in a sensible manner, the class ServiceBusHelper should be split into multiple classes, one class for handling each messaging service. It is probably a good idea to have a common, abstract base class for all of them.

supereddie:
I can spend some time to (make a start with) this issue. Initially I think it is easier to 'just move code' to separate classes without changing the implementation too much. This should make for smaller PRs that are easier to review.

I was also thinking of making a split between 'messaging actions' and 'management actions', though that might be going a bit overboard?

SeanFeldman:
> I was also thinking of making a split between 'messaging actions' and 'management actions', though that might be going a bit overboard?

Agree. Don't bother until there's a need for it.

Also, we should try to get away from "helpers". Those end up being classes with mixed concerns.

ErikMogensen:
If think it would be nice if you split the classes between 'messaging actions' and 'management actions', but it does not need to be 100%. Just where you see you can do it without too much effort.

supereddie:
I have taken a short peek that the current way the messages are send, and a lot of it is via the public `MessagingFactory` property. A lot of the sending/receiving is already outside the `ServiceBusHelper` class - it's just not in the correct place: UI controls. But it seems quite challenging to de-couple the UI controls from the specific SDK implementation for sending/receiving messages (creating queue/topic/subscription clients and such), especially with the different structure in the new SDK. 

This de-coupling could probably be part of the other issue about using the new SDK? That way this issue can be more focused on refactoring itself to make the later de-coupling easier?
Or is it better to try to do the de-coupling in this issue as well? A lot of the management actions also use/expose classes that are bound to the SDK (looking at you, `TopicDescription` and friends). These would need to be abstracted in a bunch separate data classes, together with things like `ServiceBusHelperEventArgs`...

ErikMogensen:
I suggest you go for the easiest (least difficult) path. Small steps are better.

SeanFeldman:
What @ErikMogensen said. UI coupling is an issue all over the place for SBE just because it's a traditional WinForms app that started as a utility and grew into a monster. My personal take would be whatever makes sense. And if the strategy needs to change, then you change it.

supereddie:
[...]

PR #816, opened 2024-12-03
#686: Move Queue management functions to separate class

supereddie:
First part of refactoring the ServiceBusHelper class #686. This is a fairly straight-forward 'move code to different class' to simplify the changes, and to set-up a basic structure of classes.

This part is moving Queue 'management actions' - that is, actions for Queues that need the `Microsoft.ServiceBus.NamespaceManager` to work.

### DIFF
--- a/src/Common/WindowsAzure/IServiceBusQueue.cs
+++ b/src/Common/WindowsAzure/IServiceBusQueue.cs
@@ -1,0 +1,52 @@
+﻿using Microsoft.ServiceBus.Messaging;
+using ServiceBusExplorer.Helpers;
+using ServiceBusExplorer.Utilities.Helpers;
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+#pragma warning disable IDE0130 // Namespace does not match folder structure
+namespace ServiceBusExplorer.WindowsAzure
+#pragma warning restore IDE0130 // Namespace does not match folder structure
+{
+    public interface IServiceBusQueue
+    {
+        #region Management functions
+        QueueDescription CreateQueue(QueueDescription description);
+        
+        QueueDescription CreateQueue(string path);
+        
+        Task DeleteQueue(QueueDescription queueDescription);
+        
+        Task DeleteQueue(string path);
+        
+        Task DeleteQueues(IEnumerable<string> queues);
+        
+        QueueDescription GetQueue(string path);
+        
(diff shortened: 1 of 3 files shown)


supereddie:
I'm not sure about code-style guidelines and such, with things about Resharper ignore-comments and `#region` usage for example, so any remarks are appreciated.

ErikMogensen:
> I'm not sure about code-style guidelines and such, with things about Resharper ignore-comments and `#region` usage for example, so any remarks are appreciated.

We don't have any formal recommendation for those issues and it is not consistent in the existing code. So do as you wish regarding that.

I took a quick look at the changes and it looks great. I'll take a closer look when I get time. This will be a great help when migrating to the latest ServiceBus SDK. Have you considered making one or more shared base classes for queues, topics and subscriptions? That may be a future task - if it is a good thing...

supereddie:
> I took a quick look at the changes and it looks great. I'll take a closer look when I get time. This will be a great help when migrating to the latest ServiceBus SDK.

I was wondering for the SDK migration if it would be useful to have our own custom types representing things in the ServiceBus, instead of using the SDK types. It does provide some additional abstraction and will make the switch smoother, but it can be very tricky and mistakes are easily made. In the refactoring of the ServiceBusHelper it could be one of the (last?) tasks to do.

> Have you considered making one or more shared base classes for queues, topics and subscriptions? That may be a future task - if it is a good thing...

I have thought about it, but at the moment I don't yet know what is useful to put in a base class since I haven't even looked at the code of other entity types😃When it is time, I'll probably name it something like ServiceBusEntity to make clear it is a generalized class.

src/Common/WindowsAzure/IServiceBusQueue.cs
```diff
@@ -0,0 +1,50 @@
...
+using ServiceBusExplorer.Utilities.Helpers;
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace ServiceBusExplorer.WindowsAzure
+{
+    public interface IServiceBusQueue
+    {
+        #region Management functions
```
SeanFeldman, line 12: I think it's safe to drop `#region`s in 2024. It's a relic of the past that's not needed.

supereddie: I have removed them from the new classes, and keep it in mind for future ones.

ErikMogensen (approved):
Nice!

PR #818, opened 2024-12-30
#686: Move Topic management functions to separate class

supereddie:
Part 02 of refactoring the ServiceBusHelper class #686. This is a fairly straight-forward 'move code to different class' to simplify the changes.

This part is moving Topic 'management actions' - that is, actions for Topics that need the `Microsoft.ServiceBus.NamespaceManager` to work.

In addition:
- Added a base class `ServiceBusEntity` to share some code between Queue and Topic implementations.
- Changed two properties in `ServiceBusHelper` to have a `private set` to indicate it is not set from outside the class.

### DIFF
--- a/src/Common/Helpers/ServiceBusHelper.cs
+++ b/src/Common/Helpers/ServiceBusHelper.cs
@@ -104,15 +104,10 @@ namespace ServiceBusExplorer
         private const string NotificationHubDescriptionCannotBeNull = "The notification hub description argument cannot be null.";
         private const string RuleCannotBeNull = "The rule argument cannot be null.";
         private const string PathCannotBeNull = "The path argument cannot be null or empty.";
-        private const string NewPathCannotBeNull = "The new path argument cannot be null or empty.";
         private const string NameCannotBeNull = "The name argument cannot be null or empty.";
         private const string DescriptionCannotBeNull = "The description argument cannot be null.";
         private const string ServiceBusIsDisconnected = "The application is now disconnected from any service bus namespace.";
         private const string ServiceBusIsConnected = "The application is now connected to the {0} service bus namespace.";
-        private const string TopicCreated = "The topic {0} has been successfully created.";
-        private const string TopicDeleted = "The topic {0} has been successfully deleted.";
-        private const string TopicRenamed = "The topic {0} has been successfully renamed to {1}.";
-        private const string TopicUpdated = "The topic {0} has been successfully updated.";
         private const string SubscriptionCreated = "The {0} subscription for the {1} topic has been successfully created.";
         private const string SubscriptionDeleted = "The {0} subscription for the {1} topic has been successfully deleted.";
         private const string SubscriptionUpdated = "The {0} subscription for the {1} topic has been successfully updated.";
@@ -198,6 +193,7 @@ namespace ServiceBusExplorer
         private TransportType currentTransportType;
         private ServiceBusNamespace serviceBusNamespaceInstance;
         private IServiceBusQueue serviceBusQueue;
+        private IServiceBusTopic serviceBusTopic;
         #endregion
 
         #region Private Static Fields
@@ -258,6 +254,7 @@ namespace ServiceBusExplorer
             SharedAccessKeyName = serviceBusHelper.SharedAccessKeyName;
             TransportType = serviceBusHelper.TransportType;
             serviceBusQueue = serviceBusHelper.serviceBusQueue;
+            serviceBusTopic = serviceBusHelper.serviceBusTopic;
         }
         #endregion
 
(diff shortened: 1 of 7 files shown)


src/Common/Helpers/ServiceBusHelper.cs
```diff
@@ -2103,76 +2002,34 @@ public TopicDescription CreateTopic(TopicDescription topicDescription)
...
-            throw new ApplicationException(ServiceBusIsDisconnected);
+            return serviceBusTopic.UpdateTopic(topicDescription);
         }
 
         /// <summary>
         /// Deletes all the topics in the list.
         /// <param name="topics">A list of topics to delete.</param>
         /// </summary>
-        public async Task DeleteTopics(IEnumerable<string> topics)
+        public Task DeleteTopics(IEnumerable<string> topics)
```
SeanFeldman, line 2012: I suggest we await here and don't return a `Task`. You don't gain a lot of performance by returning a task but you will reduce the changes of incorrect outcome.

src/Common/Helpers/ServiceBusHelper.cs
```diff
@@ -2103,76 +2002,34 @@ public TopicDescription CreateTopic(TopicDescription topicDescription)
...
-            await Task.WhenAll(topics.Select(DeleteTopic));
+            return serviceBusTopic.DeleteTopics(topics);
         }
 
         /// <summary>
         /// Deletes the topic described by the relative name of the service namespace base address.
         /// </summary>
         /// <param name="path">Path of the topic relative to the service namespace base address.</param>
-        public async Task DeleteTopic(string path)
+        public Task DeleteTopic(string path)
```
SeanFeldman, line 2021: same here

src/Common/Helpers/ServiceBusHelper.cs
```diff
@@ -2103,76 +2002,34 @@ public TopicDescription CreateTopic(TopicDescription topicDescription)
...
-            }
+            return serviceBusTopic.DeleteTopic(path);
         }
 
         /// <summary>
         /// Deletes the topic passed as a argument.
         /// </summary>
         /// <param name="topic">The topic to delete.</param>
-        public async Task DeleteTopic(TopicDescription topic)
+        public Task DeleteTopic(TopicDescription topic)
```
SeanFeldman, line 2030: same here

src/Common/WindowsAzure/IServiceBusQueue.cs
```diff
@@ -1,44 +1,33 @@
...
+﻿using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
+using Microsoft.ServiceBus.Messaging;
+using ServiceBusExplorer.Helpers;
 
 namespace ServiceBusExplorer.WindowsAzure
 {
-    public interface IServiceBusQueue
+    internal interface IServiceBusQueue : IServiceBusEntity
```
SeanFeldman, line 9: ❤️

src/Common/WindowsAzure/ServiceBusEntity.cs
```diff
@@ -0,0 +1,94 @@
...
+        {
+            get
+            {
+                return scheme;
+            }
+            set
+            {
+                scheme = value;
+            }
+        }
```
SeanFeldman, line 42: Any reason this couldn't be an auto-property?

SeanFeldman (changes requested):

src/Common/WindowsAzure/ServiceBusEntity.cs
supereddie: Nopoe, no reason at all. I will change it.

src/Common/Helpers/ServiceBusHelper.cs
supereddie: The main benefit I can see with returning a Task, is that you don't have to worry about `.ConfigureAwait(false)` or `.ConfigureAwait(true)` (which is default if you don't specify it) in the implementation - it can be left to the caller.

But it has been such a long time since I've done anything with Windows Forms + UI and background threads that I can't remember any 'best practices' regarding this 😆 Perhaps it is better left for a completely different issue for optimization stuff with utilizing background threads and not blocking the UI thread.

I will implement the suggestion, and see if I also did it with the previous PR for Queues and fix that at the same time.

SeanFeldman: `.ConfigureAwait()` was designed so that, by default, it caters to UI-type projects where thread affinity matters. So adding `.ConfigureAwait(true)` is the same as not having it. Only for libraries does it matter to use `.ConfigureAwait(false)`. And since this code base is all about UI, you can remove `.ConfigureAwait(true)` altogether. 

A more detailed explanation is [here](https://stackoverflow.com/questions/62500130/why-is-writing-configureawaitfalse-on-every-line-with-await-always-recommended#62505101).

SeanFeldman (approved):

PR #819, opened 2025-01-06
#686: Move Subscription management functions to separate class

supereddie:
Part 03 of refactoring the ServiceBusHelper class #686. This is a fairly straight-forward 'move code to different class' to simplify the changes.

This part is moving Subscription 'management actions' - that is, actions for Subscriptions that need the Microsoft.ServiceBus.NamespaceManager to work.

In addition:
- Add a general OnCreated() method because some Subscription events use a custom type `SubscriptionWrapper` for  `ServiceBusHelperEventArgs.EntityInstance`, which in turn is used in various UI Controls


SeanFeldman (approved):

ErikMogensen (approved):
Good

PR #822, opened 2025-01-13
#686: Move Relay management functions to separate class

supereddie:
Part 05 of refactoring the ServiceBusHelper class #686. This is a fairly straight-forward 'move code to different class' to simplify the changes.

This part is moving Relay 'management actions' - that is, actions for Relays that need the `Microsoft.ServiceBus.NamespaceManager` to work.

### DIFF
--- a/src/Common/WindowsAzure/ServiceBusRelay.cs
+++ b/src/Common/WindowsAzure/ServiceBusRelay.cs
@@ -1,0 +1,175 @@
+﻿using Microsoft.ServiceBus;
+using Microsoft.ServiceBus.Messaging;
+using ServiceBusExplorer.Enums;
+using ServiceBusExplorer.Helpers;
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace ServiceBusExplorer.WindowsAzure
+{
+    internal sealed class ServiceBusRelay : ServiceBusEntity, IServiceBusRelay
+    {
+        private const string RelayCreated = "The relay {0} has been successfully created.";
+        private const string RelayDeleted = "The relay {0} has been successfully deleted.";
+        private const string RelayUpdated = "The relay {0} has been successfully updated.";
+
+        private readonly string servicePath = string.Empty;
+
+        public ServiceBusRelay(ServiceBusNamespace serviceBusNamespace, Microsoft.ServiceBus.NamespaceManager namespaceManager)
+            : base(serviceBusNamespace, namespaceManager)
+        {
+        }
+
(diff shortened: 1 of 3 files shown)


src/Common/WindowsAzure/ServiceBusRelay.cs
```diff
@@ -0,0 +1,175 @@
...
+using ServiceBusExplorer.Helpers;
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace ServiceBusExplorer.WindowsAzure
+{
+    internal sealed class ServiceBusRelay : ServiceBusEntity, IServiceBusRelay
```
SeanFeldman, line 13: Question - would it be better to have `ServiceBusRelay` jus the a `IServiceBusRelay`, where `IServiceBusRelay` extends `IServiceBusEntity` and use composition rather than inheritance?
I.e.

```c#
class ServiceBusRelay : IServiceBusRelay, IServiceBusEntity
{
  ServiceBusRelay(IServiceBusEntity delegatedTo) {}

  // IServiceBusEntity.SomeMethod()
  public async Task SomeMethod() => await _delegatedTo.SomeMethod();
}
```

SeanFeldman: Also, if a class is `internal`, sealing it feel unnecessary

supereddie: Well, composition over inheritance would be better in the long run. However at this point I didn't want to over-complicate the instantiation of the classes and having inheritance is a bit easier to understand imo, plus the current changes are (hopefully) short-lived because of the new SDK implementation 🤞 


If needed I can change it all in one go after the moving the 'management actions' of the remaining types - EventHub and NotificationHub. Mainly because then it is clear what the various dependencies are for all the types.


About the sealing of classes: that is just a habit I have to make every class either `abstract`, or `sealed` (regardless of public/internal/private). Because in most applications I worked on there was almost never a reason to inherit from a non-abstract class, and marking it as sealed should also communicate that the class is not intended to be inherited from in the first place so it helps to prevent accidents.
Even with internal classes, sometimes one is adding a class similar to an existing one in functionality and the temptation to inherit from it is easily done and mistakes are easily made (especially with implicit casting and the `var` type) - which is why composition over inheritance is generally better.

supereddie: I have removed the sealed from this class. I will change the other classes when I'll make other changes to keep PRs focussed.

ErikMogensen (approved):
Thanks, @supereddie

PR #825, opened 2025-02-01
Set warning as error on all projects

ErikMogensen:
Set warning as error on all projects


SeanFeldman (approved):
I'm a fan of treating warnings as errors. Because they eventually become such 😀

One improvement we might apply is extract common settings into `Directory.Build.props` to remove the duplication. `TreatWarningsAsErrors` is a good candidate for that.

PR #828, opened 2025-02-25
#686: Remove unused code and make some methods private

supereddie:
Part 10 of refactoring the `ServiceBusHelper` class #686. This is a 'code cleanup' where unused code in `ServiceBusHelper` (and in the related WindowsAzure.ServiceBus* files) is removed, and a few methods are changed from `public` to `private` since there was no usage outside the class.


SeanFeldman (approved):
👍

PR #838, opened 2025-05-14
Add number of queues/topis on the root node

robionekenobi:
Add the number of queue/topics belonging to the root node of topics/queue


SeanFeldman:
Nit: when a UI change is suggested, it's recommended to add a screenshot shot of what's the expected outcome should look like.

The convention today is to show number of **messages** per node (entity). Showing a number of **entities** could be confusing, no?

robionekenobi:
Sorryx, here is the shcreenshot
image
image

SeanFeldman (approved):

ErikMogensen (approved):

PR #844, opened 2025-08-05
fix: automatically select queue or topic again after the queue/topic …

Wolf-Code:
The addition of the queue and topic count has broken the automatic queue/topic selection in the SelectEntityForm.
This change ensures it works again, as having lots of queues makes this bug quite problematic when resubmitting messages.


ErikMogensen (approved):
Nice!

ErikMogensen:
This does not work if the list of entities has been filtered. To solve this I suggest changing the copying of the nodes to copy Name -> Name instead of Name -> Text. Then you could compare using node.Name instead.

PR #849, opened 2025-08-13
Put back the the submenus for Set Status

ErikMogensen:
Put back the the submenus for Set Status again. Same issue as #760. Reused #761.


SeanFeldman (approved):
This PR emphasizes again how desperately the repo is in need of regression testing

PR #851, opened 2025-10-22
fix: Initialize ServiceBusNamespaces dictionary in helper class to resolve object reference issue

### LINKED ISSUE #841: Cannot Launch Servicebus "object reference not set to an instance of an object"
I've been trying to launch the application through the .exe file in C:\ProgramData\chocolatey\lib\servicebusexplorer but I keep getting this error.

This only happened after I installed v6.1.1

Image

SaiKanthK:
Fixes #841

### Fix: Initialize ServiceBusNamespaces Dictionary

- **Issue:** Object reference error due to uninitialized `ServiceBusNamespaces` in helper class.  
- **Resolution:** Initialized `ServiceBusNamespaces` dictionary which ensures stability and prevent runtime failures.  
- **Testing:** Verified with all existing unit tests.  
- **Result:** ✅ 42 unit test cases passed successfully.

### DIFF
--- a/src/ServiceBusExplorer/Forms/MainForm.cs
+++ b/src/ServiceBusExplorer/Forms/MainForm.cs
@@ -331,7 +331,7 @@ namespace ServiceBusExplorer.Forms
                 Keys.Control | Keys.D5
             };
 
-            foreach (var namespaceKey in serviceBusHelper?.ServiceBusNamespaces?.Keys?.OrderBy(k => k))
+            foreach (var namespaceKey in serviceBusHelper.ServiceBusNamespaces.Keys.OrderBy(k => k))
             {
                 if (serviceBusHelper.ServiceBusNamespaces[namespaceKey].UserCreated)
                 {


src/ServiceBusExplorer/Forms/MainForm.cs
```diff
@@ -331,7 +331,7 @@ private void UpdateSavedConnectionsMenu()
                 Keys.Control | Keys.D5
             };
 
-            foreach (var namespaceKey in serviceBusHelper.ServiceBusNamespaces.Keys.OrderBy(k => k))
+            foreach (var namespaceKey in serviceBusHelper?.ServiceBusNamespaces?.Keys?.OrderBy(k => k))
```
SeanFeldman: With the change,  `serviceBusHelper.ServiceBusNamespaces` is initialized and cannot be null, no? So with the `Keys` property. I.e. this change wouldn't be needed.

PR #864, opened 2026-04-02
feat: add version-bump.js for automated version management

RaoulJacobs:
## Summary

- Adds `tools/version-bump.js` — a Node.js script to bump the version across all `.csproj` files and `AssemblyInfo.cs` in one command
- Supports nested `.csproj` repos (e.g. `src/SubProject/`)
- Usage: `node tools/version-bump.js [major|minor|patch]`

## Motivation

Manual version bumping across multiple project files is error-prone. This script automates the process consistently.

## Test plan

- [ ] Run `node tools/version-bump.js patch` — verify all `.csproj` and `AssemblyInfo.cs` versions are bumped
- [ ] Run `node tools/version-bump.js minor` — verify minor version increments correctly
- [ ] Verify no unrelated files are modified


SeanFeldman:
@RaoulJacobs have you looked at the [publish action](https://github.com/paolosalvatori/ServiceBusExplorer/blob/main/.github/workflows/publish.yml)?

I don't think we need to introduce nodejs helper file when there's an already established pattern of using PowerShell.

RaoulJacobs:
Hi Sean, good point! Replaced with version-bump.ps1, consistent with the existing SetVersion.ps1 pattern in the repo. Updated in the latest commit.
________________________________

SeanFeldman:
@RaoulJacobs, I'm still not quite clear why that script is introduced. What was the need? When SBE is deployed, it's versioned  by the CI/CD scripts.

RaoulJacobs:
Hi Sean, that's a fair point. The script was added to support manual version bumping for local/fork builds where the CI/CD pipeline isn't involved. If it doesn't align with how the main project manages versioning, happy to close this PR — it was mainly useful on our end.

SeanFeldman:
Forks are free to do what they want and how they want it. When a feature from a fork can benefit other users, we're happy to take it in. Especially when it solves an existing problem, or adds a helpful feature. Like the #866 and #865 that you've contributed (thank you!). When it comes to this PR, it doesn't align with the repo's objectives, being so specific to a single user's needs. The tool is distributed already versioned and that's when it matters. For local development, not so much. If the change would align with deployment and not introduce a completely different method of versioning, I could see it working. But as a stand-alone script to be invoked manually for local compilation only, I'm going to vote 'no'.

Thank you for the other contributions. Those are valuable and appreciated.

PR #868, opened 2026-04-15
Add Microsoft Entra ID (Azure AD) authentication for Service Bus

Rogn:
Add interactive browser sign-in for Azure Service Bus namespaces using Microsoft Entra ID, addressing issue #607. Users can now connect without storing SAS secrets locally.

Features:
- Authentication selector in Connect dialog (SAS / Azure Active Directory)
- Interactive browser sign-in via MSAL with persistent token cache
- Saved Entra connections persist metadata only (endpoint, auth mode, tenant ID, transport type, entity path) - no secrets stored
- AAD support across namespace browsing, queue/topic/subscription access, Event Hub send/listen flows and queue/subscription purge flows
- Public Azure cloud only (sovereign clouds not supported)

Implementation:
- AadCredentialFactory: thread-safe credential caching shared by both old (WindowsAzure.ServiceBus) and new (Azure.Messaging.ServiceBus) SDKs
- ServiceBusNamespace: AAD connection string parser and builder
- ConnectForm: auth mode UI with field visibility toggling
- ServiceBusHelper/ServiceBusHelper2/ServiceBusPurger: AAD-aware client creation

Tests: 78 passing (3 new test classes with 22 AAD-specific tests)


src/Common/Common.csproj
```diff
@@ -21,6 +21,8 @@
 
     <PackageReference Include="Microsoft.Azure.NotificationHubs" Version="1.0.9" />
 
+    <PackageReference Include="Azure.Identity" Version="1.13.2" />
```
SeanFeldman: Is there a reason not to use the latest, 1.21.0?

src/ServiceBus/ServiceBus.csproj
```diff
@@ -12,6 +12,7 @@
 	  <TreatWarningsAsErrors>True</TreatWarningsAsErrors>
 	</PropertyGroup>
 	<ItemGroup>
+		<PackageReference Include="Azure.Identity" Version="1.13.2" />
```
SeanFeldman: Same here

SeanFeldman (commented):
Very valuable PR, @Rogn. Thank you.
A single question on my part and I'll let others review it as well.

src/ServiceBus/ServiceBus.csproj
Rogn: bumped to 1.21.0, no reason for the older version

src/Common/Common.csproj
Rogn: bumped to 1.21.0, no reason for the older version

SeanFeldman (approved):

paolosalvatori:
@claude Please review this PR and suggest improvements, if any.

paolosalvatori:
Hi @Rogn thanks for the PR 🙏 Please see Claude Code suggestions above, thanks.

SeanFeldman:
We've got help. Awesome!

Rogn:
Thanks for the thorough review. I addressed the low-risk items in this PR and kept the broader UX/threading changes for follow-up.

**Fixed in latest push (`e2f4d9c`)**
- **#1** Added a null guard in `ConnectionStringWithoutTransportType`; it now returns `null` when no SAS connection string is present.
- **#5** Removed the duplicate XML `<summary>` on `BuildAadConnectionString`.
- **#6** Narrowed `FullyQualifiedNamespace` error handling to `catch (UriFormatException)`.

I also added regression coverage for #1 and #6. The test suite is now at 80 passing tests.

**Deferred to follow-up PRs**
- **#2** `TokenCredential.GetToken` is a synchronous override, so this implementation currently uses a sync bridge. I left that unchanged here to keep this PR low-risk; happy to revisit with a broader threading-safe refactor.
- **#3** Sign-out/token clearing makes sense, but it is a new UX feature rather than a bug fix. Happy to add it in a follow-up.
- **#4** The `SemaphoreSlim` instances currently live in static app-lifetime caches. Adding disposal cleanly needs lifecycle work, so I treated that as follow-up.
- **#7** Input validation is good feedback. I would like to handle that together with a broader UX polish pass.
- **#8** Cancellation handling deserves a dedicated pass so it can be threaded through the sync call chain safely and tested properly.
- **#9** The credential creation test only checks object construction; it does not call `GetTokenAsync`, so it does not launch interactive auth.

If needed, I am happy to follow this with a more thorough UX-focused PR.

Rogn:
### Follow-up: Auth cancellation / retry fix

On closer review, **#8 (cancellation handling)** was more than a UX polish item — if the user cancelled or failed the interactive browser login, `RetryHelper` would retry up to 10 times, potentially re-opening the browser repeatedly.

**Fixed in `fde75e6`:**

1. **`RetryHelper` (all 4 overloads)** — `OperationCanceledException` is now immediately rethrown, never retried. This follows the existing pattern for `MessagingEntityAlreadyExistsException` and `CommunicationException`.

2. **`AadCredentialFactory.CreateOldSdkAuthenticationCallback`** — `AuthenticationFailedException` from the interactive browser flow is now caught and wrapped in `OperationCanceledException`, so it escapes the retry loop cleanly.

3. **5 new tests** in `RetryHelperTests.cs` verify that `OperationCanceledException` is never retried across all `RetryHelper` overloads, including the wrapped-auth-failure pattern.

**85 tests passing** (80 previous + 5 new).

ErikMogensen:
Excellent @Rogn, this is probably the most needed change for this repo. 

@claude please review PR again using the information in this thread

ErikMogensen:
@Rogn, Copilot found some issues. I removed the ones you have addressed previously so the numbers are off:


### 🟡 Medium

#### 3. Redundant token cache layered on top of MSAL

`AadCredentialFactory` maintains its own `accessTokens` dictionary with double-checked locking. `InteractiveBrowserCredential` already has an internal MSAL cache configured via `TokenCachePersistenceOptions`. The outer cache adds complexity for no gain, can serve stale tokens after MSAL has refreshed them internally, and is keyed only by `(tenantId, scopes)` — so a forced MSAL refresh won't be reflected until `AccessTokenRefreshBuffer` (5 min) expires.

#### 4. `ArgumentException` constructed with wrong overload

The parameter name is passed as the message string instead of the second argument.

```cs
// ServiceBusHelper.cs — CreateEventHubClient
throw new ArgumentException(nameof(path));  // ❌ "path" becomes the message, not the param name
```

Fix:
```cs
throw new ArgumentException("Value cannot be null or empty.", nameof(path));
```


#### 6. `CreateOldSdkAuthenticationCallback` ignores the `authorityUrl` parameter passed by the SDK

`GetAuthority(normalizedTenantId)` is always used and `authorityUrl` is silently discarded. This is correct for the common case but breaks if the old SDK calls the callback with a different authority (e.g. a sovereign cloud endpoint).

```cs
return async (authorityUrl, resource, state) =>  // authorityUrl silently discarded
{
    var token = await GetAccessTokenAsync(normalizedTenantId, GetScopes(resource), ...)
```

---

### 🔵 Low

#### 7. Static credential and token dictionaries are never purged

The six static `ConcurrentDictionary` instances in `AadCredentialFactory` grow indefinitely — one entry per tenant connected to within the process lifetime. There is no eviction, size cap, or cleanup path.

#### 8. Tests use reflection to invoke private methods and mutate static state

If `BuildCurrentConnectionString` or the private static fields are renamed, reflection calls silently return `null`. Consider marking the relevant logic as `internal` and adding `[InternalsVisibleTo("ServiceBusExplorer.Tests")]` to the assembly instead.

```cs
var method = instance.GetType().GetMethod(methodName, BindingFlags.Instance | BindingFlags.NonPublic);
SetPrivateStaticField(typeof(ConnectForm), "connectionString", null);
```

#### 9. Event handlers fired in the constructor before state is ready

`cboServiceBusNamespace_SelectedIndexChanged` and `validation_TextChanged` are called at the end of the `ConnectForm` constructor. Both read `serviceBusHelper.ServiceBusNamespaces`, which may not be fully populated when a fresh `ServiceBusHelper` is passed (as in the unit tests). The tests work around this with `ResetManualConnectionState`, which couples test setup to implementation details.

### My comments to the above ###

I am not sure if all of them needs to be fixed, so feel free to argue against them 😄 

Please also fix the issue mentioned by Claude previously:

GetNamespaceNameFromEndpoint has a bare catch (Exception)
[ServiceBusNamespace.cs:672](https://github.com/paolosalvatori/ServiceBusExplorer/pull/src/Common/Helpers/ServiceBusNamespace.cs)

Rogn:
Thanks for the thorough review, @ErikMogensen! Here's a summary of what I addressed and what I'd like to defer.

## Fixed in this push (c962e35)

**#3 — Redundant token cache removed.** Removed the custom `accessTokens` dictionary, `AccessTokenRefreshBuffer`, `IsUsableAccessToken`, and double-checked locking from `AadCredentialFactory`. MSAL's internal cache (configured via `TokenCachePersistenceOptions`) now owns token freshness. The `SemaphoreSlim` gate is kept to prevent duplicate browser prompts on concurrent first-use calls — added a comment explaining this.

**#4 — `ArgumentException` overload fixed.** Changed to `throw new ArgumentException("The path argument must not be null or whitespace.", nameof(path))` so `ParamName` is correct. Added a regression test that asserts `ex.ParamName == "path"`.

**Bare catch from prior Claude review — fixed.** Narrowed `GetNamespaceNameFromEndpoint` from `catch (Exception)` to `catch (UriFormatException)`. Added regression test with a malformed endpoint URI.

## Intentionally deferred

**#6 — `authorityUrl` parameter ignored.** This is by design. The current implementation is explicitly public-Azure-only: both the authority (`login.microsoftonline.com`) and the audience (`servicebus.azure.net`) are hard-coded. Properly honoring `authorityUrl` would require end-to-end sovereign-cloud support, not just a local callback change. Happy to tackle that as a follow-up if there's demand.

**#7 — Static dictionaries never purged.** After removing the redundant token cache, the remaining caches are low-cardinality (one entry per tenant) and process-lifetime reuse structures appropriate for a desktop app. No unbounded growth risk in practice.

**#8 — Reflection-heavy tests.** Agree on the maintainability point. `InternalsVisibleTo` already exists on the `ServiceBusExplorer` assembly but not on `Common` (where `AadCredentialFactory` lives). Switching to `internal` would be a focused testability cleanup — better as a separate follow-up than expanding this auth PR.

**#9 — Constructor event handlers.** Agree this is a design smell in `ConnectForm`, but it's a pre-existing pattern, not a regression from this PR. Also better addressed in a separate cleanup.

---

Test count: 89 (was 85). All passing.

ErikMogensen:
When entering a personal account I got:

> AADSTS500200: User account 'xxxxxxx@hotmail.com' is a personal Microsoft account. Personal Microsoft accounts are not supported for this application unless explicitly invited to an organization. Try signing out and signing back in with an organizational account. 

Is that the expected behavior?

Rogn:
@ErikMogensen Yes — that error was expected with the previous default behavior.

When Tenant ID was left blank, the AAD flow used the common endpoint, which allows personal Microsoft accounts into the browser sign-in flow. Service Bus authentication itself is organizational-only, so a personal MSA then fails with AADSTS500200 during token exchange.

I've pushed a follow-up fix in 28238d8 to tighten that default:
- blank Tenant ID now uses the organizations endpoint instead of common
- the default sign-in flow therefore accepts work/school accounts only
- personal Microsoft accounts only work if they have been invited into a Microsoft Entra tenant that has access to the namespace, in which case the tenant should be entered explicitly

I also updated the tooltip/README wording so this is clearer, and added regression coverage. Test count is now 93 passing.

ErikMogensen:
Awesome, I was able to login using a personal account when specifying the tenant id. It did not work when the tenant was blank. In that scenario I got a 3-4 login screens until it gave up. Is that something you can fix?

It did not work when trying to login to an "personal" namespace using av organizational account, but that is an edge scenario that is not critical.

I was able to login to an organization using an organizational account. However, it tried to to login to Event hubs and the other messaging services, which didn't work. One way of solving that would be to force the selection of services to only queues and topics if Entra login is selected.

Rogn:
@ErikMogensen I implemented your suggestion in `8dc9058`.

- When Entra ID auth is selected, Service Bus Explorer now scopes browsing to Service Bus queues and topics only. Subscriptions still load under topics.
- At runtime, AAD connections no longer try to probe Event Hubs, Notification Hubs, or Relay, even if those were previously selected.
- Saved AAD entries also reopen in that restricted mode.

So this change addresses the post-connect attempts against the other messaging services.

The blank-tenant + personal-account prompt churn is a separate auth-authority issue. Blank tenant now uses the `organizations` authority, but personal Microsoft accounts still need to be invited as guests and work reliably with an explicit tenant ID. I did not try to broaden that behavior further in this PR.

I also left the org-account -> personal-namespace edge case unchanged, as discussed, since that one seemed non-critical for this PR.

ErikMogensen (approved):
Excellent work! I do wish for some minor GUI changes but that is better to handle in a separate PR.

ambhatia:
Does this work for Eventhub? 
I had implemented AAD auth for both Service Bus and Eventhub in 61.3 but to a private repo. I can contribute to add the same in here. Working on this

Issue #870, opened 2026-04-20
Feature: Search/filter messages by body content and JSON properties


simonhedlund-exs:
## Problem
When investigating messages in a queue or subscription, there's no way to search or filter by message body content. The existing SQL filter only matches on system/custom properties, not the actual message payload. For queues with thousands of messages, finding a specific message requires scrolling and manually inspecting each one.

## Proposed Solution
Add the ability to filter peeked messages by body content with three filter modes:

1. **Free-text search** — case-insensitive substring match on the deserialized message body
2. **JSON path filter** — select a JSON path (auto-discovered from loaded messages) and filter by value at that path
3. **SQL property filter expression** — the existing filter, integrated into the same dialog

### UI
- An inline search box ("Search body...") in the Message List header for quick filtering
- A filter dialog (accessible via the existing filter icon) with all three modes combined
- Same filtering available for Dead-letter messages, with an additional "Dead Letter Reason" filter
- Filters applied in-memory on already-peeked messages (no additional Service Bus calls)

### Scope
- `HandleQueueControl` and `HandleSubscriptionControl`
- Both Messages and Dead-letter tabs

I have a working implementation in my fork and would be happy to submit a PR.

PR #871, opened 2026-04-20
Add body content filtering to message and deadletter views

simonhedlund-exs:
Implements #870 

- Add inline search boxes for filtering messages by body content and deadletter reason
- Implement body content filtering with free text and JSON path matching
- Add BodyFilterForm dialog for configuring body filters (free text, JSON path, SQL property filter)
- Support case-sensitive toggle and auto-discovered JSON paths
- Cache message bodies for filtering performance
- Apply to both queue (HandleQueueControl) and subscription (HandleSubscriptionControl) views


SeanFeldman:
@claude

simonhedlund-exs:
@SeanFeldman 
Addressed all review feedback in the latest commit. Key changes:

Replaced custom JSON parser with [Newtonsoft.Json.Linq](vscode-file://vscode-app/c:/Users/SimonHedlund/AppData/Local/Programs/Microsoft%20VS%20Code/560a9dba96/resources/app/out/vs/code/electron-browser/workbench/workbench.html)
Extracted shared code to [MessageBodyFilterHelper](vscode-file://vscode-app/c:/Users/SimonHedlund/AppData/Local/Programs/Microsoft%20VS%20Code/560a9dba96/resources/app/out/vs/code/electron-browser/workbench/workbench.html)
Fixed cache/disposal bugs, reverted unrelated PeekBatch change
Build passes, all existing tests green.

ErikMogensen:
Wonderful, just what I needed 😄! I checked out your branch and started filtering. However, the debugger raised the following issue;

> Managed Debugging Assistant 'ContextSwitchDeadlock' 
>   Message=Managed Debugging Assistant 'ContextSwitchDeadlock' : 'The CLR has been unable to transition from COM context 0x960c2bd0 to COM context 0x960c2aa8 for 60 seconds. The thread that owns the destination context/apartment is most likely either doing a non pumping wait or processing a very long running operation without pumping Windows messages. This situation generally has a negative performance impact and may even lead to the application becoming non responsive or memory usage accumulating continually over time. To avoid this problem, all single threaded apartment (STA) threads should use pumping wait primitives (such as CoWaitForMultipleHandles) and routinely pump messages during long running operations.'

It was kind enough to offer a fix which I accepted. I'll try to create a PR to your branch with the proposed fix in a couple of hours.

SeanFeldman:
@claude review again. See what could cause ContextSwitchDeadlock.

ErikMogensen:
Hmmm, that's a different approach that I got. In my case it put the work in an asynchronous task to offload the UI thread. If the fix Claude suggested in this thread works I think that's a better solution. @simonhedlund-exs, wanna try the [the fix this link from above](https://claude.ai/code?q=In%20HandleQueueControl.cs%20and%20HandleSubscriptionControl.cs%20on%20branch%20feature%2Fbody-search-filter%2C%20move%20EnsureMessageBodyCache%20calls%20to%20immediately%20after%20messageBindingList%20and%20deadletterBindingList%20are%20assigned%20in%20GetMessages%2FGetDeadLetterMessages%20methods.%20Remove%20EnsureMessageBodyCache%20calls%20from%20FilterMessages%2C%20FilterDeadletters%2C%20and%20the%20filter%20button%20click%20handlers.%20The%20cache%20will%20be%20pre-populated%20during%20the%20existing%20load%20phase%20instead%20of%20blocking%20the%20UI%20thread%20during%20filtering.&repo=paolosalvatori/ServiceBusExplorer)?

simonhedlund-exs:
@ErikMogensen 
 I went with the approach Claude suggested — moved [EnsureMessageBodyCache](vscode-file://vscode-app/c:/Users/SimonHedlund/AppData/Local/Programs/Microsoft%20VS%20Code/560a9dba96/resources/app/out/vs/code/electron-browser/workbench/workbench.html) to immediately after [messageBindingList](vscode-file://vscode-app/c:/Users/SimonHedlund/AppData/Local/Programs/Microsoft%20VS%20Code/560a9dba96/resources/app/out/vs/code/electron-browser/workbench/workbench.html)/[deadletterBindingList](vscode-file://vscode-app/c:/Users/SimonHedlund/AppData/Local/Programs/Microsoft%20VS%20Code/560a9dba96/resources/app/out/vs/code/electron-browser/workbench/workbench.html) are assigned during the loading phase, and removed the calls from [FilterMessages()](vscode-file://vscode-app/c:/Users/SimonHedlund/AppData/Local/Programs/Microsoft%20VS%20Code/560a9dba96/resources/app/out/vs/code/electron-browser/workbench/workbench.html), [FilterDeadletters()](vscode-file://vscode-app/c:/Users/SimonHedlund/AppData/Local/Programs/Microsoft%20VS%20Code/560a9dba96/resources/app/out/vs/code/electron-browser/workbench/workbench.html), and the filter button click handlers.

This way the cache is pre-populated while the WaitCursor is already showing during load, and filtering just reads from the Dictionary — no WCF calls, no blocking, no deadlock.

Pushed in commit 6a57c5d. Let me know if you still see the ContextSwitchDeadlock after this change!

ErikMogensen:
@simonhedlund-exs, I ran the PR through Copilot and it found the following issues:

**High**

1. Body cache is populated unconditionally at load time, even without an active filter. Every peek/receive now calls GetMessageText on every message (WCF XML decode, potential gzip decompress, BrokeredMessage.Clone() even if the user never touches the body filter. For 500 messages this adds significant overhead to every refresh.

2. Log flooding on every keypress is still present. writeToLog(string.Format(BodyFilterAppliedMessage, ...) fires inside FilterMessages/FilterDeadletters on every debounce tick. Typing a 6-character search term produces 6 log entries.

**Medium**

3. BodyFilterForm uses hard-coded absolute pixel coordinates throughout. All controls placed with new Point(x, y) — breaks on high-DPI displays. FormBorderStyle.FixedDialog prevents resizing, but buttons are anchored Bottom | Right while their Location is fixed, so they'll drift incorrectly if the DPI scaling changes the client area.

4. Tag ordering bug in placeholder restore. In AddInlineSearchBox (and AddDeadLetterReasonSearchBox), the LostFocus handler sets txt.Text = "Search body..." before txt.Tag = true. This fires TextChanged with Tag still false, triggering the debounce timer and an unnecessary filter cycle.

5. Bare catch blocks. ExtractJsonPaths and GetJsonValueAtPath use catch { } and catch { return null; } which silently swallow OutOfMemoryException and ThreadAbortException. Should be catch (JsonException).

**Low**

6. WasCleared is dead code. Set in BtnClear_Click but never read by either calling control.

7. PopulateDeadLetterReasonCombo is misleadingly named. It only subscribes a TextChanged event on a TextBox — no combo, no population.

8. Redundant Close() calls. Setting DialogResult on a modal form already closes it; the explicit Close() in BtnOk_Click and BtnClear_Click is a no-op.

9. ~110 lines of code duplicated verbatim across HandleQueueControl and HandleSubscriptionControl (filter fields, timer wiring, Apply*Filter methods, HasAnyDeadletterBodyFilter,  additions).

**My comments to the above points**
You can ignore number 2. I believe the reason for  number 3 is that you ran [Visual Studio as a DPI-aware process](https://github.com/paolosalvatori/ServiceBusExplorer#development-environment). 

It is hard getting a PR approved these days...

ErikMogensen:
@simonhedlund-exs, I used your updated PR today and it solved my problem 😃. Thanks 🙏🏻 !

ErikMogensen:
I suggest to solve the first one by reading the message bodies during the first search and then storing them in a cache.

ErikMogensen:
I realize now that the maximum queue size is 80 GB which would break most machines if it was stored in the cache. That complicates the caching 😢

simonhedlund-exs:
@ErikMogensen Went through all the points and pushed a new commit.

The biggest change is the cache strategy — you were right that pre-populating on every load was wasteful. Bodies now get decoded on a background thread only when a filter is actually applied, and the filter just re-runs automatically when it's done. Added a version counter so a slow background task from an old load can't corrupt the current view.

Also fixed the LostFocus placeholder issue (was setting Text before Tag, which would fire the debounce timer on blur), swapped out the bare catch {} blocks, removed the redundant Close() calls in the dialog, and cleaned up the dead WasCleared property.

Let me know if anything looks off!

ErikMogensen:
My Copilot says:

## Review: `feature/body-search-filter` (commit `99a8b5a`)

This covers the latest commit — _perf: Cache message bodies lazily on first filter use_ — and includes issues carried forward from earlier commits that remain unresolved.

---

### 1. Dictionary race condition — background thread writes while UI thread reads 🔴

`capturedCache` in the `Task.Run` lambda and `messageBodyCache` on the UI thread are the **same `Dictionary<BrokeredMessage, string>` instance**. `Dictionary<>` is not thread-safe. If the user opens the filter dialog while a background populate is still running, the dialog-open handler immediately calls the synchronous `EnsureMessageBodyCache` on the UI thread — writing to the same dictionary simultaneously from two threads. This can silently corrupt the dictionary's internal state.

```csharp
// Task.Run background thread — writes capturedCache (same ref as messageBodyCache):
capturedCache[msg] = serviceBusHelper.GetMessageText(msg, useAscii, out _);

// UI thread in filter button click — also writes messageBodyCache concurrently:
if (!_messageCacheReady)
{
    MessageBodyFilterHelper.EnsureMessageBodyCache(messageBindingList, messageBodyCache, ...);
```

**Suggestion:** Use `ConcurrentDictionary<BrokeredMessage, string>`, or set `_messageCachePopulating = true` before starting the task and skip the synchronous fallback while population is in progress.

---

### 2. Filter dialog blocks the UI thread with no feedback 🔴

The filter button click handler falls through to a synchronous `EnsureMessageBodyCache` call when `_messageCacheReady` is false. This blocks the UI thread for as long as it takes to decode all message bodies, with no cursor change or progress indicator. The previous commit had `Cursor.Current = Cursors.WaitCursor` which was removed here.

```csharp
if (!_messageCacheReady)
{
    // Blocks UI thread — no cursor, no progress bar, no cancellation
    MessageBodyFilterHelper.EnsureMessageBodyCache(messageBindingList, messageBodyCache,
        msg => serviceBusHelper.GetMessageText(msg, MainForm.SingletonMainForm.UseAscii, out _), writeToLog);
    _messageCachePopulating = false;
    _messageCacheReady = true;
}
```

**Suggestion:** If the async cache is still running, disable the filter button and re-enable it in the `onReady` callback, or show a brief status message and return early.

---

### 3. `catch (Exception)` in background task silently swallows all errors 🟡

The `catch` block inside `Task.Run` catches `Exception` but does nothing except store an empty string. This means `OutOfMemoryException`, corruption, and unexpected errors are silently hidden with no log output — unlike the synchronous `EnsureMessageBodyCache` in `MessageBodyFilterHelper` which at least calls `writeToLog`.

```csharp
catch (Exception)
{
    capturedCache[msg] = string.Empty;
}
```

**Suggestion:** Catch only expected types (`InvalidOperationException`, `IOException`), or at minimum log the failure:

```csharp
catch (Exception ex)
{
    writeToLog($"[BodyFilter] Failed to decode body for {msg.MessageId}: {ex.Message}");
    capturedCache[msg] = string.Empty;
}
```

---

### 4. Log flooding on every debounce tick 🟡

`writeToLog(string.Format(BodyFilterAppliedMessage, filteredList.Count))` fires inside `FilterMessages` and `FilterDeadletters` on every debounce timer tick. Typing a six-character search term produces six log entries. This pollutes the log and can cause noticeable UI lag on large message sets.

```csharp
if (hasBodyFilter)
{
    filteredList = filteredList.Where(...).ToList();
    writeToLog(string.Format(BodyFilterAppliedMessage, filteredList.Count)); // fires on every keypress
}
```

**Suggestion:** Only log when the filter is applied via the explicit "Apply" button, not during inline live search.

---

### 5. `BodyFilterForm` uses hard-coded absolute pixel coordinates throughout 🟡

All controls are positioned with `new Point(x, y)` and sized with `new Size(w, h)`. This is inconsistent with the rest of the codebase and breaks layout on high-DPI displays (125%, 150%, 200% scaling).

```csharp
grpPropertyFilter = new GroupBox { Location = new Point(12, 12), Size = new Size(480, 60), ... };
txtPropertyFilter  = new TextBox  { Location = new Point(90, 22), Size = new Size(375, 20), ... };
btnOk              = new Button   { Location = new Point(310, 345), Size = new Size(85, 30), ... };
```

**Suggestion:** Use a `.resx` / Designer file with `AutoScaleMode = AutoScaleMode.Dpi`, or use layout panels (`FlowLayoutPanel`, `TableLayoutPanel`) so the form scales correctly.

---

### 6. `catch (Exception)` in `ExtractJsonPaths` and `GetJsonValueAtPath` 🟡

Both static helpers catch `Exception` broadly instead of catching only `JsonException` (or `Newtonsoft.Json.JsonReaderException`). This masks bugs and can hide fatal CLR exceptions.

```csharp
catch (Exception)
{
    // Not valid JSON, skip
}
```

**Suggestion:** Use `catch (Newtonsoft.Json.JsonException)` to limit the scope to expected parse failures.

---

### 7. `PopulateDeadLetterReasonCombo` is misleadingly named 🟢

The method does not populate anything. It only subscribes a `TextChanged` event on a `TextBox`. The name implies it fills a combo box.

```csharp
private void PopulateDeadLetterReasonCombo()
{
    if (txtDeadLetterReasonFilter == null) return;
    txtDeadLetterReasonFilter.TextChanged -= OnDeadLetterReasonTextChanged;
    txtDeadLetterReasonFilter.TextChanged += OnDeadLetterReasonTextChanged;
}
```

**Suggestion:** Rename to `SubscribeDeadLetterReasonFilter` or `WireDeadLetterReasonFilter`.

---

### 8. ~100 lines of code duplicated verbatim across two controls 🟢

`HandleQueueControl` and `HandleSubscriptionControl` contain identical copies of:
- `EnsureMessageBodyCacheAsync` / `EnsureDeadletterBodyCacheAsync`
- `ApplyInlineMessagesFilter` / `ApplyInlineDeadletterFilter` / `ApplyDeadLetterReasonFilter`
- `OnMessagesInlineFilterChanged` / `OnDeadletterInlineFilterChanged`
- All filter state fields (`_messageCacheReady`, `messagesBodyFreeTextFilter`, etc.)

Any future bug fix or behaviour change must be applied in both places.

**Suggestion:** Extract a `MessageBodyFilterState` class (holding the cache, version counter, and filter values) and a shared `MessageBodyFilterController` that owns the async cache methods, then compose it into both controls.

## My comments to the above
@simonhedlund-exs, you can ignore number 4 and 8, but please fix the others. 

There is also the issue of searching through 80 GB of messages which would freeze most computers. I suggest you check the size of the cache and stop adding to it if it is larger than 2 GB.

simonhedlund-exs:
Hi @ErikMogensen , thanks for the feedback! Fixed everything in 949c4d3:

- `ConcurrentDictionary` for both caches (queue + subscription)
- `writeToLog` moved out of `Task.Run` — collected in a local list and flushed in `ContinueWith` on the UI thread
- `TryAdd` instead of `ContainsKey` + indexer
- 2 GB cap with a log message when hit
- `WaitCursor` on the sync populate calls
- `AutoScaleMode.Dpi` on `BodyFilterForm`
- `catch (JsonException)` instead of `catch (Exception)`
- Renamed `PopulateDeadLetterReasonCombo` → `WireDeadLetterReasonFilter`

Let me know if anything else needs changing!

ErikMogensen:
Thank you @simonhedlund-exs, I ran it through my local Copilot and got the following:


# Code Review: `feature/body-search-filter` 
## Issues

### 1. Significant code duplication — `EnsureMessageBodyCacheAsync` / `EnsureDeadletterBodyCacheAsync`

Both methods in `HandleQueueControl` (and again in `HandleSubscriptionControl`) are ~80 lines of near-identical code. The only differences are which binding list, cache, flags, and version counter are used. This results in roughly **320 lines of duplication** across the two controls.

**Suggestion:** Extract into a reusable class (e.g., `MessageBodyCache`) that encapsulates the state and async population logic, then use instances for messages and dead-letters:

```csharp
// replaces 4 copies of EnsureMessageBodyCacheAsync
private readonly MessageBodyCache _messageCache = new MessageBodyCache();
private readonly MessageBodyCache _deadletterCache = new MessageBodyCache();
```

---

### 2. `Font` objects leak GDI handles — `MessageBodyFilterHelper`

`AddInlineSearchBox` and `AddDeadLetterReasonSearchBox` both create `new Font("Microsoft Sans Serif", 8.25F)`. WinForms `TextBox` does not take ownership of assigned fonts; these objects are never disposed.

```csharp
// Leaks a Font GDI handle:
Font = new Font("Microsoft Sans Serif", 8.25F),
```

**Suggestion:** Use `parent.Font` or the system default font instead, or dispose the font when the parent control is disposed.

---

### 3. `BodyFilterForm` not using WinForms designer — inconsistent with codebase

The new form builds its UI entirely in `InitializeForm()`. All other forms in this project use `.Designer.cs` files. This makes it impossible to visually edit in the designer and harder to maintain.

**Suggestion:** Move to the standard WinForms designer pattern (a `BodyFilterForm.Designer.cs` file with `InitializeComponent()`).

---

### 4. JSON path with dot-notation silently fails for keys containing dots

`GetJsonValueAtPath` splits the path on `'.'`, so any JSON key that itself contains a dot (e.g. `"data.v1"`) will silently produce no match. Users have no indication that their path failed vs. simply not matching.

**Suggestion:** Either document this limitation clearly in the UI tooltip/hint text, or use `SelectToken` from Json.NET which supports proper JSONPath syntax natively:

```csharp
// Handles nested paths, arrays, and keys with dots correctly:
var result = jobj.SelectToken(path);
```

---

### 5. `EnsureMessageBodyCache` (sync) blocks the UI thread

The synchronous cache pre-population in `MessageBodyFilterHelper.EnsureMessageBodyCache` is called with `Cursor.Current = Cursors.WaitCursor` but has no timeout or max-size guard — unlike the async version which caps at 2 GB and uses a background thread. For large message sets, opening the filter dialog can freeze the UI.

**Suggestion:** Use consistent behavior — either always use the async path (populate in the background before showing the dialog), or add the same 2 GB cap and early-exit to the sync path.

---

### 6. `_messageCachePopulating` flag is not guarded against re-entrant calls

If `FilterMessages()` is called while `_messageCachePopulating = true`, it returns without scheduling a new callback. If the filter state changes again (e.g. user clears the inline filter then retypes), the re-entrant call is silently dropped. The version counter protects against stale completions, but the dropped re-entrant call is unrecovered.

**Suggestion:** When returning `false` from `EnsureMessageBodyCacheAsync` because a population is already in-flight, store the latest `onReady` callback and invoke it (instead of the original one) when population completes.

---

### 7. Minor: `BtnClear_Click` sets `DialogResult` directly

```csharp
private void BtnClear_Click(object sender, EventArgs e)
{
    ...
    DialogResult = DialogResult.OK;
}
```

Setting `this.DialogResult` closes a modal form (intended behavior), but it bypasses `AcceptButton`/`CancelButton` semantics and is non-obvious.

**Suggestion:** Assign `this.DialogResult` and call `this.Close()` explicitly, or wire the button's `DialogResult` property in the constructor and only set the output properties in `BtnOk_Click`.

---

## What's Done Well

- Async body cache with version counter correctly prevents stale completions after a message list refresh.
- Debounced timers (400ms) for inline search prevent excessive re-filtering on each keystroke.
- `WireDeadLetterReasonFilter` uses unsubscribe-then-subscribe pattern to prevent double-wiring.
- 2 GB cache cap in the async path prevents OOM on large namespaces.
- `JsonException` is caught specifically rather than swallowing all exceptions.
- `ConcurrentDictionary` is used correctly for the cache shared between UI and background threads.

## My comments to the review above
You can ignore the first one about code duplication, I can create a separate PR that. Please fix the other ones though unless you object...

ErikMogensen:
@simonhedlund-exs, I see that you made some commits. Let us know when it is ready for review again, thanks.

simonhedlund-exs:
Hi @ErikMogensen 

Thanks for the detailed review and patience! I've addressed all the issues from your latest feedback. Here's what's been fixed:

**✅ Issue #2 - Font GDI handle leak:**
Removed `Font = new Font("Microsoft Sans Serif", 8.25F)` from both `AddInlineSearchBox` and `AddDeadLetterReasonSearchBox`. TextBoxes now inherit parent font.

**✅ Issue #4 - JSON path dot-notation:**
Replaced the manual path-splitting logic with `SelectToken()` from Newtonsoft.Json.Linq. This properly handles keys containing dots, nested arrays, and all valid JSONPath expressions.

**✅ Issue #5 - Sync EnsureMessageBodyCache blocking UI:**
Added the same 2GB cache limit and byte-counting to the sync method that the async version has. Now both paths have consistent safeguards and the sync version also shows `Cursor.WaitCursor` during population.

**✅ Issue #6 - Re-entrancy callback handling:**
Implemented `_latestMessagesOnReady` and `_latestDeadletterOnReady` fields. When a re-entrant call occurs during cache population, the latest callback is stored and invoked when population completes instead of dropping the call.

**✅ Issue #7 - BtnClear_Click DialogResult:**
Made the dialog closure explicit (though I see this is actually quite minor since setting DialogResult on a modal form closes it anyway).

**✅ Additional thread-safety improvement (just now):**
Changed the sync `EnsureMessageBodyCache` to use `cache.TryAdd()` instead of `ContainsKey` + indexer assignment for consistency with the async implementation. Also updated the method signature to explicitly use `ConcurrentDictionary` for clarity.

**Regarding Issue #3 (WinForms Designer):**
Since you mentioned ignoring the code duplication, I figured this might be similar priority? The form works correctly with `AutoScaleMode.Dpi` (same as `CreateEventGridSubscriptionForm` uses). Happy to convert it to Designer format if that's important, but wanted to check first.

Build passes with 0 errors. All existing tests green. Ready for review! 🙂

Best,
Simon

simonhedlund-exs:
Hi, Any updates here? :)

ErikMogensen (commented):
Great work @simonhedlund-exs! If you fix the fix the three issues mentioned by Copilot in the last review and add logging if the data cannot fit in the cache, I'll approve it. Given that a partitioned queue can contain 80 GB, filtering such a queue will make SBE appear to have frozen. 

Since SBE runs on a great variety of hardware a possible future enhancement would be to check available memory and use an algorithm for deciding cache size. But that does not need to be a part of this PR.

PR #874, opened 2026-04-24
Entra auth for eventhub

ambhatia:
This PR adds Entra ID (Azure Active Directory) authentication support for **Event Hub namespaces**. Currently, AAD auth works with Service Bus namespaces, but connecting to an Event Hub namespace with AAD isn't supported. This change makes it possible to use AAD authentication seamlessly with both Service Bus and Event Hub namespaces.

## Existing implementation
- Event Hub entity loading was explicitly blocked for AAD connections in both `ConnectForm` and `MainForm`
- `EventHubClient.CreateWithAzureActiveDirectory()` internally hard-codes a Service Bus resource in its auth callback, which fails for Event Hub namespaces

### Scope probe + fallback (`ServiceBusHelper.Connect()`)
- On AAD connect, tries the Service Bus scope first via `GetQueues()`
- If it fails with an auth error (indicating an Event Hub namespace), gracefully falls back to the Event Hub scope (`https://eventhubs.azure.net`)
- Sets an `IsEventHubNamespace` property accordingly

### Audience-aware credential caching (`AadCredentialFactory`)
- Added `EventHubsAudience` constant
- Added audience-parameterized overloads for `GetTokenProvider()` and `GetAuthenticationCallback()`
- Updated cache keys to `tenantId|audience` format to avoid cross-scope cache hits

### Event Hub client creation (`ServiceBusHelper.CreateEventHubClient()`)
- Uses `MessagingFactory` with explicit AMQP transport and AAD token provider instead of `EventHubClient.CreateWithAzureActiveDirectory()` (which has the hard-coded SB resource issue mentioned above)
- The factory is cached during `Connect()` and reused across calls to avoid per-call AMQP connection overhead
- Resilient to stale connections: checks `IsClosed` before use and auto-recreates on connection failures with a single retry

### UI changes
- `ConnectForm.SelectedEntities` now includes Event Hub entities for AAD connections
- Removed the `!isAad` guard from Event Hub loading in `MainForm`

### Reliability improvement
- Added `UnauthorizedAccessException` to non-retriable exceptions in all 4 `RetryHelper` methods to avoid unnecessary retry loops on auth failures

## Testing

- 17 unit tests pass (12 existing + 5 new)
- Manually tested: AAD connect to Event Hub namespace, list Event Hubs, Send Events, Consumer Group listener

I've done my best to keep the changes minimal and focused. Happy to address any feedback or make adjustments — thank you for your time reviewing this! 🙏


ErikMogensen:
Thank you @ambhatia! 

Let's start with: @claude please review this pr

ErikMogensen:
@ambhatia, please fix the issues found above. Then I'll run the PR though my local Copilot.

ambhatia:
Thanks Eric, have addressed the comments.

1. PR Review Comment Fixes:
------------------------------------------------------
- Guarded loadQueues/loadTopics with !IsEventHubNamespace so they don't run on Event Hub namespaces.
- Gated loadEventHubs so it only runs when IsEventHubNamespace is true.
- Replaced "Unauthorized" string match with "401"/"40100" status code checks for auth error detection.
- Fixed TOCTOU race on eventHubMessagingFactory by capturing to a local variable before use.

2. Scope Probe Optimization
----------------------------
- Replaced GetQueues() with QueueExists("$__scope_probe__$") for scope detection.
- Single lightweight REST call instead of enumerating all queues.
- Returns false on Service Bus (valid scope), throws UnauthorizedAccessException on Event Hub (triggers fallback).

3. Reconnect State Reset
-------------------------
- Reset IsEventHubNamespace = false and eventHubMessagingFactory = null at start of Connect().
- Prevents stale state from a previous connection leaking into a new one.

4. EntityPath Bug Fix
---------------------
- When EntityPath is set, load only the specified entity instead of enumerating all entities.
- ServiceBusQueue/ServiceBusTopic: try-catch MessagingException when EntityPath refers to a wrong entity type, return empty list.
- MainForm: use GetEventHubAsync(entityPath) instead of GetEventHubsAsync() when EntityPath is set.
- Defensive try-catch for Event Hub EntityPath: catches MessagingException, returns empty list.

5. Logging Improvements
-----------------------
- Connection log now says "event hub namespace" or "service bus namespace" based on detected scope.
- Added warning log when EntityPath doesn't match any event hub in the namespace.

6. RetryHelper Hardening
-------------------------
- Added UnauthorizedAccessException catch-and-rethrow in all 4 retry methods.
- Prevents masking auth failures during scope probe.

ErikMogensen:
# Branch Review: Entra ID (AAD) Event Hub Support using my local Copilot

**Branch vs `main`** — 11 commits, 9 files changed, +269 / -34


## Purpose

Add Entra ID (AAD) authentication support for Event Hub namespaces, with auto-detection of namespace type at connect time.

---

## Per-File Review

### `AadCredentialFactory.cs`

**Good:**
- `ServiceBusAudience` and `EventHubsAudience` are now `public const` — clean, discoverable.
- New overloads `CreateOldSdkAuthenticationCallback(tenantId, audience)` and `CreateOldSdkTokenProvider(tenantId, audience)` keep backward compatibility via zero-arg overloads.
- Cache key uses `{tenantId}|{audience}` — correctly isolates different audience tokens.

**Potential concern:**
- The inline comment explains why `normalizedAudience` overrides the SDK-supplied `resource` in the callback. This is intentional and well-documented, but it's a subtle behavior break from the SDK contract. Worth calling out in the PR description for future maintainers.

---

### `RetryHelper.cs`

**Good:**
- `UnauthorizedAccessException` is now rethrown immediately in all 4 retry loops (before `MessagingException`/`ServerBusyException` catch blocks). Auth failures should not be retried.
- Pattern is consistent across all overloads.

---

### `ServiceBusHelper.cs`

**Good:**
- `IsEventHubNamespace` and `EntityPath` properties added cleanly.
- Scope probe using `QueueExists("$__scope_probe__$")` is lightweight (single REST call, no enumeration).
- `eventHubMessagingFactory` is cached and reused; stale factory is recreated with a single retry.
- Reset of `IsEventHubNamespace = false` and `eventHubMessagingFactory = null` on reconnect prevents state leakage.
- `MessagingFactory = null` for EH namespaces is correct (no queues/topics).

**Concern — thread safety (Medium):**

```csharp
var factory = eventHubMessagingFactory;
if (factory == null || factory.IsClosed)
{
    factory = CreateEventHubMessagingFactory();
    eventHubMessagingFactory = factory;
}
```

This is a TOCTOU race — two threads could both see `IsClosed == true` and both create a new factory. The old factory would be leaked (never closed/disposed). Not critical for a desktop app, but worth a `lock` or `Interlocked` pattern if `CreateEventHubClient` is called concurrently (e.g., parallel EH partition loading).

**Minor:**
- The log message format change (`{0} {1} namespace`) is fine; `IsEventHubNamespace ? "event hub" : "service bus"` reads clearly.

---

### `ServiceBusQueue.cs` / `ServiceBusTopic.cs`

**Good:**
- Catching `MessagingException` on the `EntityPath` path and returning an empty list is the right defensive approach when an EH namespace has an `EntityPath` that refers to an event hub, not a queue/topic.

**Concern — overly broad catch (Low):**
- Catching all `MessagingException` is broad — it would also swallow transient errors (e.g., network blip) on a legitimate SB namespace with an `EntityPath`. Consider narrowing to `MessagingEntityNotFoundException` or checking `ex.IsTransient` to avoid masking real failures.

**Concern — silent failure (Low):**
- The catch block returns an empty list with no log output, making it hard to diagnose unexpected failures. At minimum, a `WriteToLog` call in the catch block would help.

---

### `ConnectForm.cs`

**Good:**
- AAD connections now include `EventHubEntities` in `SelectedEntities` — correct since namespace type isn't known until connect time.
- The display label update (`Queue, Topic (Subscription), Event Hub`) is minor UI polish consistent with the change.

---

### `MainForm.cs`

**Good:**
- `loadQueues` and `loadTopics` now guard on `!serviceBusHelper.IsEventHubNamespace` — correctly skips SB entities for EH namespaces.
- `loadEventHubs` now allows EH loading for AAD when `IsEventHubNamespace == true`.
- `EntityPath` branching: loads a single EH by path or falls back to `GetEventHubsAsync()`.

**Minor:**
- The `MessagingException` catch on the single-hub fetch logs the error and falls back to an empty list — acceptable, though a more specific exception type would be preferable (same concern as `ServiceBusQueue.cs`/`ServiceBusTopic.cs`).

---

### `ConnectFormAadTests.cs` / `ServiceBusHelperAadTests.cs`

**Good:**
- Tests updated to reflect the new `EventHubEntities` in the AAD entity list.
- New tests cover: audience isolation, caching, callback isolation, correct audience value, and scope string format.
- The test setup for `CreateEventHubClient` now properly primes `aadTokenProvider`, `namespaceUri`, and `eventHubMessagingFactory` via reflection — correctly mirrors the new `Connect()` flow.

---

## Issues Summary

| Severity | File | Issue |
|---|---|---|
| Medium | `ServiceBusHelper.cs` | `eventHubMessagingFactory` assignment is not thread-safe — potential factory leak under concurrent `CreateEventHubClient` calls |
| Low | `ServiceBusQueue.cs`, `ServiceBusTopic.cs` | `MessagingException` catch is too broad — could swallow transient errors on SB namespaces with `EntityPath` |
| Low | `ServiceBusQueue.cs`, `ServiceBusTopic.cs` | Catch block returns empty list silently — no log output, difficult to diagnose |

---

## Verdict

The core logic is sound and the commit history shows good iterative refinement (scope probe optimization, reconnect state reset, factory caching/resilience). The thread-safety concern in `CreateEventHubClient` is the most notable issue to address before merging. The broad `MessagingException` catches are low risk in practice but worth tightening for correctness.

## My comments to the above
Too bad that Claude suggested a solution that wasn't thread safe. Please fix that and the catches.

ambhatia:
addressed the comments and added few more improvements:

1. **Thread-safe Event Hub factory** — Added `eventHubFactoryLock`; narrowed catch filters to `!ex.IsTransient`; added logging in catch blocks.

2. **Fixed duplicate log messages** — Moved EntityPath logging from utility classes to `ShowEntities()` caller so it fires only once.

3. **Narrowed scope probe** — Introduced `IsAudienceMismatchException()` to distinguish audience-mismatch (switch to EH) from permission-denied (stay on SB). 5 new unit tests.

4. **Factory leak fix** — Added `ReplaceEventHubFactory()` helper that `Close()`s old factory before assigning new one. Applied at all 3 replacement sites.

5. **TOCTOU fix** — Wrapped remaining `ReplaceEventHubFactory()` calls in `lock(eventHubFactoryLock)` for consistency.

ErikMogensen (approved):
Wonderful!

PR #876, opened 2026-04-28
fix: Ctrl+F in message text box no longer hijacked by TreeView filter (#873)

### LINKED ISSUE #873: v6.2.0 removed the ability to search in the Message Text box
The global Ctrl+F _Filter queues/topics_ feature introduced in #865 removed the ability to search text in the Message Text box.

I used to be able to focus the Message Text box while investigating a Deadletter message and press Ctrl+F to search the message body for text. However, after updating to v6.2.0, I've lost this ability. Now when I press Ctrl+F, it focuses the _Filter queues/topics_ box (as designed).

Would it be possible to add an exception so that whenever the Message Text box is focused, Ctrl+F opens the old text search box instead of focusing the _Filter queues/topics_ search box?

RaoulJacobs:
## Summary

Fixes #873 — `Ctrl+F` pressed while the message body (`FastColoredTextBox`) has focus no longer redirects to the TreeView filter box.

## Root cause

`ProcessCmdKey` in `MainForm` intercepted `Ctrl+F` unconditionally, including when a `FastColoredTextBox` (message text, deadletter text) had focus.

## Fix

Added a `GetFocusedControl` helper that walks the `ContainerControl` hierarchy to find the actual focused control. If it is a `FastColoredTextBox`, the key event is passed through to the control's native handler.

```csharp
protected override bool ProcessCmdKey(ref Message msg, Keys keyData)
{
    if (keyData == (Keys.Control | Keys.F))
    {
        if (GetFocusedControl(this) is FastColoredTextBoxNS.FastColoredTextBox)
            return base.ProcessCmdKey(ref msg, keyData);

        filterTreeViewTextBox.Focus();
        filterTreeViewTextBox.SelectAll();
        return true;
    }
    return base.ProcessCmdKey(ref msg, keyData);
}

private static Control GetFocusedControl(Control control)
{
    if (control is ContainerControl cc && cc.ActiveControl != null)
        return GetFocusedControl(cc.ActiveControl);
    return control;
}
```

## Test plan

- [x] Ctrl+F with Message Text box focused → stays in text box (native find)
- [x] Ctrl+F with Deadletter Text box focused → stays in text box
- [x] Ctrl+F with focus elsewhere (tree, properties grid) → focuses TreeView filter box


RaoulJacobs:
Closing — branch had too many commits. Will reopen with a single fix commit.

PR #877, opened 2026-04-28
Ctrl+F in message text box no longer hijacked by TreeView filter (#873)

### LINKED ISSUE #873: v6.2.0 removed the ability to search in the Message Text box

RaoulJacobs:
## Summary

Fixes #873 — `Ctrl+F` pressed while the message body (`FastColoredTextBox`) has focus no longer redirects to the TreeView filter box.

## Root cause

`ProcessCmdKey` in `MainForm` intercepted `Ctrl+F` unconditionally, including when a `FastColoredTextBox` (message text, deadletter text) had focus.

## Fix

Added a `GetFocusedControl` helper that walks the `ContainerControl` hierarchy to find the actual focused control. If it is a `FastColoredTextBox`, the key event is passed through to the control's native handler.

```csharp
protected override bool ProcessCmdKey(ref Message msg, Keys keyData)
{
    if (keyData == (Keys.Control | Keys.F))
    {
        if (GetFocusedControl(this) is FastColoredTextBoxNS.FastColoredTextBox)
            return base.ProcessCmdKey(ref msg, keyData);

        filterTreeViewTextBox.Focus();
        filterTreeViewTextBox.SelectAll();
        return true;
    }
    return base.ProcessCmdKey(ref msg, keyData);
}

private static Control GetFocusedControl(Control control)
{
    if (control is ContainerControl cc && cc.ActiveControl != null)
        return GetFocusedControl(cc.ActiveControl);
    return control;
}
```

## Test plan

- [x] Ctrl+F with Message Text box focused → stays in text box (native find)
- [x] Ctrl+F with Deadletter Text box focused → stays in text box
- [x] Ctrl+F with focus elsewhere (tree, properties grid) → focuses TreeView filter box

### DIFF
--- a/src/ServiceBusExplorer/Forms/MainForm.cs
+++ b/src/ServiceBusExplorer/Forms/MainForm.cs
@@ -7748,11 +7748,22 @@ namespace ServiceBusExplorer.Forms
         {
             if (keyData == (Keys.Control | Keys.F))
             {
+                if (GetFocusedControl(this) is FastColoredTextBoxNS.FastColoredTextBox)
+                {
+                    return base.ProcessCmdKey(ref msg, keyData);
+                }
                 filterTreeViewTextBox.Focus();
                 filterTreeViewTextBox.SelectAll();
                 return true;
             }
             return base.ProcessCmdKey(ref msg, keyData);
+        }
+
+        private static Control GetFocusedControl(Control control)
+        {
+            if (control is ContainerControl cc && cc.ActiveControl != null)
+                return GetFocusedControl(cc.ActiveControl);
+            return control;
         }
 
         #endregion


ErikMogensen:
Thank you @RaoulJacobs !

@claude, please check this pr

ErikMogensen (approved):
Thanks!

paolosalvatori (commented):
## Review summary

**Overall assessment:** Correct, minimal, well-scoped fix for #873. I verified the mechanics: `ProcessCmdKey` only bails out to the base handler when the actually-focused control is a `FastColoredTextBox`, so the control's native find dialog receives Ctrl+F, while every other focus target keeps the existing redirect to the TreeView filter box. The `ContainerControl.ActiveControl` recursion is the standard WinForms pattern and terminates correctly. The exact `keyData == (Keys.Control | Keys.F)` match means Ctrl+Shift+F and friends are unaffected. Build and unit tests pass on CI.

**Key risks:** Essentially none. The behavior change is confined to the case the PR intends to change (FCTB focused). The affected FCTB instances hosted inside `MainForm` (queue/subscription message and dead-letter text boxes, test/listener controls) all sit under `UserControl`s, which are `ContainerControl`s, so the focus walk reaches them.

**Status of prior review conversations:**
- No inline review threads exist on this PR.
- @ErikMogensen approved on 2026-04-28.
- The claude[bot] issue-comment review (2026-04-28) raised two optional points that were never replied to or addressed:
  1. *Missing `using FastColoredTextBoxNS;` directive* — still valid as a cosmetic nit; re-raised inline below. Not blocking.
  2. *"A plain `Panel` between two `ContainerControl`s would break the recursion"* — **obsolete/incorrect**: `Panel`/`GroupBox`/`TabPage` do not scope `ActiveControl`, so a `ContainerControl`'s `ActiveControl` skips straight past them to the focused child or the next nested container. The recursion is unaffected; no action needed and no P/Invoke required.
- Housekeeping: the `@claude review once` run requested on 2026-07-17 errored out without posting a review (actions run 29563780676).

**Recommendation:** Approve / merge. The single inline note is cosmetic.

src/ServiceBusExplorer/Forms/MainForm.cs
```diff
@@ -7748,13 +7748,24 @@ protected override bool ProcessCmdKey(ref Message msg, Keys keyData)
         {
             if (keyData == (Keys.Control | Keys.F))
             {
+                if (GetFocusedControl(this) is FastColoredTextBoxNS.FastColoredTextBox)
```
paolosalvatori, line 7751: **nit** — `MainForm.cs` has no `using FastColoredTextBoxNS;`, so the type is qualified inline here. Other files that reference this type (`MessageForm.cs`, `TabControlHelper.cs`) import the namespace and use the short name. Adding the `using` and writing `is FastColoredTextBox` would match the rest of the codebase. Purely cosmetic — fine to merge as-is. (Carried over from the earlier claude[bot] pass, where it was never picked up.)

PR #879, opened 2026-04-28
Force node to v24 and bug fixes in the pipelines

ErikMogensen:
Forced the actions to run to node v24.
Some bug fixes in the pipelines

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -20,6 +20,7 @@ on:
 env:
   SOLUTION_FILE_PATH: .\src
   BUILD_CONFIGURATION: Release
+  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
 
 jobs:
   build_test_publish:
--- a/.github/workflows/handle-tag.yml
+++ b/.github/workflows/handle-tag.yml
@@ -6,6 +6,9 @@ on:
       - '[0-9]+.[0-9]+.[0-9]+*'
   workflow_dispatch:
 
+env:
+  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
+
 jobs:
   check-tag-branch:
     # This job is based on replies in https://github.community/t/how-to-create-filter-on-both-tag-and-branch/16936/6
@@ -15,7 +18,7 @@ jobs:
       id: tag-commit-hash
       run: |
         hash=${{ github.sha }}
-        echo "{name}=tag-hash::${hash}" >> $GITHUB_OUTPUT
+        echo "tag-hash=${hash}" >> $GITHUB_OUTPUT
         echo $hash
 
     - name: Checkout main
@@ -27,7 +30,7 @@ jobs:
       id: main-commit-hash
       run: |
         hash=$(git log -n1 --format=format:"%H")
-        echo "{name}=main-hash::${hash}" >> $GITHUB_OUTPUT
+        echo "main-hash=${hash}" >> $GITHUB_OUTPUT
         echo $hash
 
     - name: Verify tag commit matches main commit - exit if they don't match
@@ -37,6 +40,7 @@ jobs:
         exit 1
 
   get-version:
+    needs: [check-tag-branch]
     runs-on: windows-2022
     steps:
     - name: Get version from tag
--- a/.github/workflows/handle-tag.yml
+++ b/.github/workflows/handle-tag.yml
@@ -14,30 +14,22 @@ jobs:
     # This job is based on replies in https://github.community/t/how-to-create-filter-on-both-tag-and-branch/16936/6
     runs-on: ubuntu-latest
     steps:
-    - name: Get tag commit hash
-      id: tag-commit-hash
-      run: |
-        hash=${{ github.sha }}
-        echo "tag-hash=${hash}" >> $GITHUB_OUTPUT
-        echo $hash
-
-    - name: Checkout main
+    - name: Checkout repository history
       uses: actions/checkout@v4
       with:
-        ref: main
+        fetch-depth: 0
 
-    - name: Get latest main commit hash
-      id: main-commit-hash
+    - name: Verify tag commit is contained in main
       run: |
-        hash=$(git log -n1 --format=format:"%H")
-        echo "main-hash=${hash}" >> $GITHUB_OUTPUT
-        echo $hash
-
-    - name: Verify tag commit matches main commit - exit if they don't match
-      if: steps.tag-commit-hash.outputs.tag-hash != steps.main-commit-hash.outputs.main-hash
-      run: |
-        echo "Tag was not on the main branch. Exiting."
-        exit 1
+        # Peel annotated tags to their target commit and verify that commit is reachable from main.
+        # This intentionally enforces "contained in main" semantics instead of exact main HEAD equality.
+        tag_commit=$(git rev-parse "$GITHUB_SHA^{commit}")
+        if git merge-base --is-ancestor "$tag_commit" origin/main; then
+          echo "Tag commit is contained in main."
+        else
+          echo "Tag commit is not contained in main. Exiting."
+          exit 1
+        fi
 
   get-version:
     needs: [check-tag-branch]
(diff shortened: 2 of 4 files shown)


ErikMogensen:
@claude, please review this pr

.github/workflows/handle-tag.yml
```diff
@@ -15,7 +18,7 @@ jobs:
       id: tag-commit-hash
       run: |
         hash=${{ github.sha }}
-        echo "{name}=tag-hash::${hash}" >> $GITHUB_OUTPUT
+        echo "tag-hash=${hash}" >> $GITHUB_OUTPUT
```
paolosalvatori: **minor** — this fix does more than tidy syntax: with the old broken format both step outputs were empty, so the verify step's `if:` never fired and every tag sailed through. This PR activates the gate for the first time. Two semantics worth confirming are what you want before the next release:

1. **Annotated tags:** on a `push` to `refs/tags/*`, `github.sha` can resolve to the *tag object* rather than the peeled commit for annotated tags, which would make this comparison fail even when the tag points at main HEAD. Lightweight tags (GitHub UI / `gh release create`) are unaffected. Worth one test with an annotated tag if anyone tags from the CLI.
2. **Exact-HEAD match:** the gate requires the tag to equal the tip of `main` *at run time* — tagging an older commit on main, or a commit racing onto main right after tagging, now fails the pipeline.

If you'd rather have "tag is contained in main" semantics (which also sidesteps the annotated-tag issue), checkout with `fetch-depth: 0` and use:

```bash
git merge-base --is-ancestor "$(git rev-parse "$GITHUB_SHA^{commit}")" origin/main
```

The current strictness may well be intentional — flagging it because the behavior visibly changes with this PR.

.github/workflows/publish.yml
```diff
@@ -80,9 +83,13 @@ jobs:
     - name: Create the release
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}  
-      continue-on-error: true
       run: |
-        gh release create ${{ env.RELEASE_VERSION }}  --draft --generate-notes
+        $releaseExists = gh release view ${{ env.RELEASE_VERSION }} 2>&1
```
paolosalvatori: **nit** — `$releaseExists` is assigned but never read; only `$LASTEXITCODE` matters. Discarding the output says what's intended:

```suggestion
        gh release view ${{ env.RELEASE_VERSION }} *> $null
```

.github/workflows/publish.yml
```diff
@@ -80,9 +83,13 @@ jobs:
...
+        if ($LASTEXITCODE -ne 0) {
+          gh release create ${{ env.RELEASE_VERSION }} --draft --generate-notes
```
paolosalvatori, line 89: **minor** — the existence check is the right replacement for `continue-on-error: true` (which also masked auth/API failures). But the very scenario this enables — re-running the workflow after a partial failure — still dies one step later: `gh release upload` (the "Upload build output zipped" step below) fails on assets that were already uploaded. Adding `--clobber` there makes the whole workflow idempotent end-to-end:

```powershell
gh release upload ${{ env.RELEASE_VERSION }} $env:ZipFilename $env:NupkgFilename --clobber
```

(Also noted in the earlier claude[bot] pass; still open.)

.github/workflows/build-test.yml
```diff
@@ -20,6 +20,7 @@ on:
 env:
   SOLUTION_FILE_PATH: .\src
   BUILD_CONFIGURATION: Release
+  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
```
paolosalvatori, line 23: **nit / informational** — `FORCE_JAVASCRIPT_ACTIONS_TO_NODE24` is the documented opt-in and workflow-level `env:` is the right scope; applied consistently across all four workflows. Two notes for the record: (1) it forces *every* JS action onto Node 24, including any action still bundled for Node 20 — the actions used here (`checkout@v4`, `cache@v4`, `claude-code-action@v1`) are fine with that; (2) it's transitional — once Node 24 is the runner default this becomes a no-op and the four `env:` blocks can be deleted. If you want to get ahead of it without the flag, bumping `actions/checkout` to its Node 24-native major (v5) does the same for the checkout steps.

paolosalvatori (commented):
## Review summary

**Overall assessment:** Small, correct, and a genuine reliability improvement to the release pipeline. The important change is the `$GITHUB_OUTPUT` fix in `handle-tag.yml`: the old `{name}=key::value` lines defined an output literally named `{name}`, so `tag-hash`/`main-hash` were always empty, the verify step's `if:` never fired, and the "tag must be on main" gate silently passed everything. This PR makes that gate live for the first time — see the inline note for two semantics to be aware of before the next release. `needs: [check-tag-branch]` correctly serializes the gate ahead of `get-version` (and transitively the build/publish jobs). The idempotent release creation is the right replacement for `continue-on-error: true`, which used to swallow *all* errors including auth/API failures.

**Key risks:**
- The newly-activated tag gate compares the tag SHA to the exact current `main` HEAD and may behave surprisingly with annotated tags (inline comment on `handle-tag.yml`).
- Re-running the publish workflow now gets past release creation but can still fail at `gh release upload` on duplicate assets (inline comment; `--clobber` closes the gap).

**Status of prior review conversations:**
- No inline review threads exist on this PR.
- Copilot review (2026-04-28): reviewed all 4 files, generated no comments.
- claude[bot] issue-comment review (2026-04-29) recommended approve and raised two minor points that were never replied to or addressed in code: (1) `$releaseExists` assigned but unused, (2) duplicate-asset failure on re-run of the upload step. Both still valid — re-raised as inline comments here so they can be tracked to closure.
- Housekeeping: the `@claude review once` run requested on 2026-07-17 errored out without posting (actions run 29563775453).

**Recommendation:** Approve / merge, ideally after adding `--clobber` to the upload step and giving the annotated-tag caveat a quick thought. Nothing blocking.

.github/workflows/build-test.yml
ErikMogensen: I interpret this as it is fine as it is.

.github/workflows/handle-tag.yml
ErikMogensen: I fixed this.

paolosalvatori (approved):
Hi @ErikMogensen 

Thanks a lot for this one and for catching the `$GITHUB_OUTPUT` problem in particular. That check had been silently broken for who knows how long, so this is a genuinely valuable fix, not just housekeeping.

I approve this PR, but please take a look at the comments I left this morning and make sure to implement them. To be explicit about what I'm asking:

1. **Add `--clobber` to the `gh release upload` step** in `publish.yml`. Your existence check now lets a failed run be re-run past the create step, but the re-run will still die on assets that were already uploaded. `--clobber` closes that last gap and makes the whole workflow safely re-runnable:

   ```powershell
   gh release upload ${{ env.RELEASE_VERSION }} $env:ZipFilename $env:NupkgFilename --clobber
   ```

2. **Drop the unused `$releaseExists` variable** — only `$LASTEXITCODE` is actually read. Something like `gh release view ${{ env.RELEASE_VERSION }} *> $null` says what's intended.

3. **Before we cut the next release, let's double-check the tag gate together.** Your fix makes the "tag must be on main" check live for the first time, and it has two sharp edges: it requires the tag to match the *exact* current HEAD of main (tagging an older commit will now fail), and annotated tags may always fail because `github.sha` can resolve to the tag object rather than the commit. If we only ever use lightweight tags we're fine, but I'd rather we verify than find out during a release. If you prefer "tag is contained in main" semantics, the `git merge-base --is-ancestor` variant I put in the review comment handles both issues.

Nothing blocking on the Node 24 changes — that's all fine, and we can delete the `FORCE_JAVASCRIPT_ACTIONS_TO_NODE24` blocks once it becomes the runner default.

Thanks again for keeping the pipelines healthy!

Paolo

ErikMogensen:
Thanks @paolosalvatori, I updated the branch hopefully fixing 2 and 3. The issue  

> 1. Add --clobber to the gh release upload step in publish.yml. Your existence check now lets a failed run be re-run past the create step, but the re-run will still die on assets that were already uploaded. --clobber closes that last gap and makes the whole workflow safely re-runnable:

I think it is better to not allow overwriting a release. In case you want to overwrite a release you can delete it.

Issue #286, opened 2018-12-22
Cross platform support


regevbr:
Hi,

I moved to Ubuntu and I have to keep a windows vm just to use your great tool. 
Can you please release a linux release as well?

regevbr:
I managed to complie it using mono after a few hour of debugging.
When I try to connect it throws many exceptions due to some issues with compatibility in linux of the azure sdk.
Suggest fix in the web is to use https://github.com/ppatierno/azuresblite which is compatible.

SeanFeldman:
Related to #284.

The new official ASB client would help here, but the project is not there yet.

knakkert:
Also for mac would be great!

Burnett2k:
Hi Guys, this tool is really awesome and I wanted to also comment on how great it would be if we could get a mac supported version out there! I've been trying to avoid getting a windows VM for a long time, but I might have to just for this tool because it makes debugging so much faster!

paolosalvatori:
Thanks @Burnett2k for the feedback, we are considering to migrate the tool to .NET Core, the problem is that the UI is built using Windows Forms that is currently not supported by .NET Core, and I have no idea if it will ever be in the future :)

epomatti:
We use Ubuntu and Mint, and would be great to have this tool available cross-platform.

SeanFeldman:
@epomatti I'll echo what Paolo has said. The tool was not designed to be a cross-platform tool. The effort to get it there cannot be undermined. Not to mention some major restructuring work that is already taking place.

Personally, I'd look to use an alternative for now while Azure Portal is working on adding some features and this project gets on the newer dependencies. Have you looked at the tools that can solve your problem today? [Cerulean](https://www.cerebrata.com/) is one of those. And it might be even free. And if it's not, if you need to use ASB and operate on Linux, perhaps it's worth shelling out a few coins to get what you need for the job. An alternative would be to actively contribute to this OSS and help implement what's needed to make it x-plat.

PS: I might lose my collaborator status by suggesting Cerulean, but we need to keep things in perspective. If your job depends on a tool you should weigh the options and not tie all your hopes on a volunteer-driven OSS. Cheers.

grahambunce:
[...]

PR #882, opened 2026-05-25
Avalonia added

Meeshka:
## Summary

This PR adds an initial cross-platform Avalonia desktop implementation for Service Bus Explorer alongside the existing WinForms app.

### Added

- New `ServiceBusExplorer.Avalonia` project targeting `.NET 8`.
- New `Common.Core` shared service layer for cross-platform Service Bus operations.
- Avalonia-based main window, connection dialog, entity tree, entity detail view, and message workflows.
- SAS and Microsoft Entra ID connection profile support.
- Connection string parsing and JSON-backed saved connections.
- Queue, topic, and subscription browsing.
- Message peek and send workflows.
- Dead-letter message browsing and purge support.
- Queue settings editing.
- Output logging in the Avalonia UI.
- macOS `.app` bundle metadata, build script, and documentation.

### Changed

- Updated the solution to include the new Avalonia and shared core projects.
- Adjusted Event Grid and existing WinForms code paths for compatibility with the shared implementation.

### Build

```text
dotnet build ServiceBusExplorer.sln --configuration Release --no-restore
Result: succeeded.
Warnings observed:
•
Common.Core/Services/ModernMessagingService.cs(227,19): nullable dereference warning.
•
MinimumRecommendedRules.ruleset was not found for the WinForms project.


SeanFeldman:
@Meeshka, thank you for your PR. 

Trying to understand what issue you're tackling here. Would this be #286? If 'yes', are you planning to take it beyond the initial stage? Would you like to share your thoughts on how it should be approached (phased conversion, re-write, etc.)? 

The PR is still in draft, so I don't want to jump the gun. Moving away from WinForms to an alternative such as Avalonia would truly work if the end result is no longer coupled to .NET Framework 4.x, and instead is on the modern .NET (Core). Feel free to share your thoughts.

Meeshka:
Hi

I feel really sorry. Actually I forked the solution and was going to make this pull request to the fork, not touching the original repo.
I have no idea how it landed to the main repository. Probably when copying the local files I may have left the original .git folder

In short I am a cloud dev that constantly work with all types of Azure resources. And swithing to macOS became a pain to work with ServiceBus.

So I finally had some time to use your ideal app and enhance it with cross platform Avalonia to support MacOS. I think that there were some requests from mac users who wanted some good tool. So this solution will definitely be good for them. It has all mostly used features for now. I can add the .sh that publishes the app for macOS to the repo.

I think that switching to .NET core is always a good idea, since .NET fw is end of life on 11.2026.

Regarding the full switching to Avalonia both in Windows and macOS, I can't recommend. Avalonia doesn't bring that functionality as winforms do.

I will definitely continue to add features in case of need. But frankly as for now I am fully satisfied with the result.

Feel free to cancel this PR if you think it's ambiguous or doesn't fit, I will switch to fork if it is ok.
If you decide to keep it, i can try to be a good support for it 😉

Kind regards,
Mikhail

--
Sent from Mail ( https://mail.ru/ ) app for Android

Monday, 25 May 2026, 08:18pm +03:00 from Sean Feldman ***@***.*** :

SeanFeldman:
The PR didn't cause any issues. But it did raise some curiosity.
Specifically, how it helps anyone who's not on Windows if the Avalonia project is producing `WinExe` output.

> I think that switching to .NET core is always a good idea, since .NET fw is end of life on 11.2026.

There's [no EOL date](https://learn.microsoft.com/en-us/lifecycle/products/microsoft-net-framework). .NET 4.x is tied to the operating system IIRC.

Meeshka:
Hi

Good question. In this case "WinExe" is a bit misleading by name, but it does not mean that the Avalonia project only produces a Windows executable.

For .NET GUI projects, "OutputType=WinExe" mainly means “build this as a windowed application instead of a console application”, so no console window is attached/opened. The actual target platform is controlled by the target framework and, more importantly, by the runtime identifier used during publish.

For example, on macOS I can publish the Avalonia project to package the result into a ".app" bundle. That is what my publish script currently does, and the produced app runs on my MacBook.

That said, I agree the "WinExe" name can look confusing. I can add a short note to the README or the project file comments explaining that platform-specific publishing is done through the RID, for example "dotnet publish -r osx-arm64", and that "WinExe" here only indicates a GUI app rather than a console app.

Kind regards,
Mikhail

--
Sent from Mail ( https://mail.ru/ ) app for Android

Monday, 25 May 2026, 09:45pm +03:00 from Sean Feldman ***@***.*** :

PR #888, opened 2026-07-18
In debug version inform user do not ask for new version

robionekenobi:
In main Form when in debug mode, it always ask for to update

image

After

image

### DIFF
--- a/src/ServiceBusExplorer/Forms/MainForm.cs
+++ b/src/ServiceBusExplorer/Forms/MainForm.cs
@@ -414,6 +414,11 @@ namespace ServiceBusExplorer.Forms
 
         void DisplayNewVersionInformation()
         {
+#if DEBUG
+            linkLabelNewVersionAvailable.Visible = true;
+            linkLabelNewVersionAvailable.Enabled = false;
+            linkLabelNewVersionAvailable.Text = $"Debug Version";
+#else
             if (!VersionProvider.IsLatestVersion(out var releaseInfo, WriteToLog))
             {
                 linkLabelNewVersionAvailable.Visible = true;
@@ -423,6 +428,7 @@ namespace ServiceBusExplorer.Forms
             {
                 linkLabelNewVersionAvailable.Visible = false;
             }
+#endif
         }
 
         private void UpdateSavedConnectionsMenu()
@@ -490,7 +496,7 @@ namespace ServiceBusExplorer.Forms
             argumentName = argument;
             argumentValue = value;
         }
-        #endregion
+#endregion
 
         #region Event Handlers
         private void duplicateSubscriptionMenuItem_Click(object sender, EventArgs e)


SeanFeldman:
I'm not sure what problem does it solve... The fact that it shows there's a new version? In that case, I'd assume it would be more helpful not to show anything rather than "Debug version". Whoever runs the tool from an IDE is well aware of that, no?

robionekenobi:
Yes, i agree, but i'm also having a link to start the debug version, so i always have the latest changees without the ide, and I know then the date of the latest change. Sure, either show nothing, but i didn't see the problem if i display that it is a debug version and when it was last changed

ErikMogensen (approved):
Thanks! 

Another scenario this is useful is when running a debug version and a released version at the same time to compare new functionality.

src/ServiceBusExplorer/Forms/MainForm.cs
```diff
@@ -414,6 +414,11 @@ private async void RefreshDashboard()
 
         void DisplayNewVersionInformation()
         {
+#if DEBUG
```
paolosalvatori, line 417: **MEDIUM | Gate on the version stamp, not on the build configuration**

**Description:** The stale prompt is caused by the version stamp, not by the Debug configuration: `.github/workflows/build-test.yml:45-48` injects `-p:FileVersion=<release-version>`, so a build made outside CI keeps the `.csproj` placeholder (`<AssemblyVersion>1.0.0.1</AssemblyVersion>`, the same value `src/ServiceBusExplorer.Tests/FileVersionShould.cs:16` treats as the unstamped default) and `VersionProvider.IsLatestVersion` compares it against the latest release, currently 6.2.0. A local `dotnet build -c Release` therefore still shows "New Version 6.2.0 is available", and because `#if DEBUG` removes the only two reachable calls into `IsLatestVersion` (this one and, once the link is disabled, `NewVersionAvailableForm.cs:82`), the update-check path stops being exercised in the configuration `TESTING.md` tells contributors to build. A runtime check on the stamp covers both configurations and keeps one code path.

**Rules applied:** General best practice. Evidence: `.github/workflows/build-test.yml:45-48`, `src/ServiceBusExplorer/ServiceBusExplorer.csproj:10`, `src/ServiceBusExplorer.Tests/FileVersionShould.cs:16-20`, `src/Common/Helpers/VersionProvider.cs:45-53`.

The fix spans lines outside this diff, so no one-click suggestion. Replace the `#if DEBUG` / `#else` pair here (and drop the `#endif` at L431) with:

```csharp
            // CI stamps the real version via -p:FileVersion (build-test.yml); a build made outside CI
            // keeps the .csproj placeholder, so every published release looks newer. Errs toward showing
            // the marker: an unstamped build is never one the user downloaded.
            if (VersionProvider.IsUnstampedBuild())
            {
                linkLabelNewVersionAvailable.Visible = true;
                linkLabelNewVersionAvailable.Enabled = false;
                linkLabelNewVersionAvailable.Text = "Local Build";
            }
            else if (!VersionProvider.IsLatestVersion(out var releaseInfo, WriteToLog))
```

and add the helper to `src/Common/Helpers/VersionProvider.cs`, next to `IsLatestVersion`, which already reads the same `FileVersionInfo`:

```csharp
        // Matches <AssemblyVersion> in ServiceBusExplorer.csproj: the value a build carries when CI
        // has not passed -p:FileVersion.
        static readonly Version UnstampedBuildVersion = new Version(1, 0, 0);

        public static bool IsUnstampedBuild()
        {
            var versionInfo = FileVersionInfo.GetVersionInfo(Assembly.GetExecutingAssembly().Location);
            return new Version(versionInfo.FileMajorPart, versionInfo.FileMinorPart, versionInfo.FileBuildPart)
                == UnstampedBuildVersion;
        }
```

src/ServiceBusExplorer/Forms/MainForm.cs
```diff
@@ -414,6 +414,11 @@ private async void RefreshDashboard()
...
+            linkLabelNewVersionAvailable.Visible = true;
+            linkLabelNewVersionAvailable.Enabled = false;
+            linkLabelNewVersionAvailable.Text = $"Debug Version";
```
paolosalvatori, line 420: **LOW | Disabling the link makes `NewVersionAvailableForm` unreachable in Debug builds**

**Description:** `Enabled = false` suppresses `LinkClicked`, and `MainForm.cs:7633-7639` is the only place `NewVersionAvailableForm` is constructed, so in a Debug build that dialog (latest release notes plus the zip download link) becomes dead UI where it was previously always reachable. The loss is small because Help > About still reports the running version (`AboutForm.cs:217`), so keeping it disabled is a defensible call; if on-demand access matters, dropping the one line preserves it and the label still reads as a build marker.

**Rules applied:** General best practice. Evidence: `src/ServiceBusExplorer/Forms/MainForm.cs:7633-7639` (sole construction site), `src/ServiceBusExplorer/Forms/NewVersionAvailableForm.cs:79-95`, `src/ServiceBusExplorer/Forms/AboutForm.cs:217`.

```suggestion
            linkLabelNewVersionAvailable.Visible = true;
            linkLabelNewVersionAvailable.Text = "Debug Version";
```

src/ServiceBusExplorer/Forms/MainForm.cs
```diff
@@ -490,7 +496,7 @@ public MainForm(string argument, string value, string logMessage)
             argumentName = argument;
             argumentValue = value;
         }
-        #endregion
+#endregion
```
paolosalvatori, line 499: **LOW | Unrelated de-indentation leaves the only misaligned region directive in the file**

**Description:** This hunk is unrelated to the PR's purpose and looks like an accidental IDE reformat. It closes `#region Public Constructor` (L269, indented 8 spaces), and after the change it is the only class-level region directive in this 7,700-line file sitting at column 0: 27 of them use 8 spaces, and the other 4 unindented ones are the file-level `Copyright` and `Using Directives` regions.

**Rules applied:** General best practice (keep a focused change focused).

```suggestion
        #endregion
```

paolosalvatori (approved):
Reviewed `src/ServiceBusExplorer/Forms/MainForm.cs` at `c3166c9` (+7 / -1, 1 file). Three inline findings: 1 MEDIUM, 2 LOW, no HIGH.

## 1. Rules applied

Read in full before reviewing: `localstack-pro-azure/CLAUDE.md`; `.claude/rules/azure/common/{coding-style,hooks,patterns,security,testing}.md`; `.claude/rules/azure/python/{cloud-pipeline,coding-style,hooks,patterns,security,testing}.md`. None was missing or unreadable. No bicep/terraform/shell rules loaded: the PR changes one `.cs` file.

**Scope conflict, and how I resolved it.** Every one of those rule files carries a `paths:` front-matter block scoping it to `localstack-pro-azure/**` (the Python ones narrow further to `**/*.py` / `**/*.pyi`). This PR is in a different repository, in C#, and is not an Azure emulator, so none of those rules has authority here and no finding cites one. Findings cite "General best practice" plus concrete evidence from this repository. The Azure parity checks are not applicable either: the change touches no Azure API, resource provider, error code, or SDK behaviour, so there is no service parity gaps checklist.

## 2. Existing comments

@ErikMogensen approved and noted a second use case (running a debug and a released build side by side). `copilot-pull-request-reviewer[bot]` recommended approval with zero comments. @SeanFeldman questioned the value and suggested showing nothing rather than a "Debug version" label; @robionekenobi replied that they launch the locally built binary from a shortcut outside the IDE and want an on-screen marker. None of that overlaps the findings below, and I have not repeated any of it.

## 3. Findings

Posted inline. Summary:

| Severity | Line | Finding |
|---|---|---|
| MEDIUM | L417 | Gate on the version stamp, not on the build configuration |
| LOW | L418-L420 | Disabling the link makes `NewVersionAvailableForm` unreachable in Debug builds |
| LOW | L499 | Unrelated de-indentation leaves the only misaligned region directive in the file |

## 4. Clean files

None. `src/ServiceBusExplorer/Forms/MainForm.cs` is the only changed file and it carries all three findings.

## 5. Verdict

**Approve.** The change is correct and safe as written: `DEBUG` is defined only by the Debug configuration (no `DefineConstants` override in `ServiceBusExplorer.csproj` or `src/Directory.Build.props`, and no configuration-conditional version property), CI builds `-c Release` and `publish.yml` packages from `bin\Release`, so shipped artifacts keep the existing behaviour unchanged. No HIGH finding; the three above are structural and cosmetic.

## Fit with the tool

It fits. Service Bus Explorer has always been a pragmatic operator tool, and a corner-of-the-window marker that stops a locally built binary from nagging about an update it cannot install is squarely in that register: it costs one screenful of code, changes nothing for users, and removes a blocking synchronous GitHub call (`VersionProvider.cs:47`, `GetAwaiter().GetResult()` on the UI thread during `MainForm` construction) from the developer's startup path.

On the objection that showing nothing would be better than showing "Debug version": the author's counter is the stronger one, and the side-by-side scenario reinforces it. Once you run the built binary from a shortcut rather than from the IDE, the window itself has to tell you which build it is, and the label costs nothing to a user who will never see it.

Worth noting that the label does not carry what was actually asked for in the discussion ("I know then the date of the latest change"), and it cannot: an unstamped build reports version 1.0.0, so printing the version would show a meaningless placeholder. If the build date is genuinely wanted, that is `File.GetLastWriteTime(Assembly.GetExecutingAssembly().Location)` and a separate change.

PR #895, opened 2026-09-13
Add bussin to the list of Service Bus tools

sgebb:
Add Bussin to the list of alternatives. It offers something a bit different from the others on the list. It's not a full Service Bus management tool, it focuses on message operations. It's also a browser-based PWA, so no installation is required.


ErikMogensen (approved):

SeanFeldman:
Thank you, @sgebb.

